PR #6, opened 2020-01-12
Logo proposal

cpehle:
As we've discussed, I've come up with a [logo](https://github.com/norse/norse/blob/logo-proposal/logo.svg). I like it for the following reasons:

- It incorporates "braiding" suggesting the relationship to both nordic mythology and to how forward and backward pass are intertwined temporally.
- It incorporates both a common notation for lambda abstraction the backslash \ and suggests the nabla differentiation operator \/.

Drawbacks: 
- Doesn't work easily for monochrome / grayscale


Jegp:
Would you be ok to merge that @cpehle?

cpehle:
Well, I think there is potentially a problem with the svg version I've posted, because it doesn't embed the font (Futura). So it might be necessary to convert it to a path first. The one you've uploaded looks horribly broken.

PR #12, opened 2020-01-22
Fixed #11

### LINKED ISSUE #11: Wrong ConvNet model parameter
Bug in ConvNet model parameter where the model uses the LIFFeedForward cell with `model` constructor parameter rather than `method` parameter

Jegp:
Added `__init__.py` to manifest `module` as an importable package
Fixed a bug in model initialisation


cpehle:
LGTM

PR #13, opened 2020-02-09
Fix doc

Jegp:
Fixes related to comments from the HBP summit

* Fixed dependencies of tutorial notebooks
* Completed (ran) tutorial notebooks
* Added documentation and more meaningful variable names to tutorial notebooks
* Added a spike plot for the front page readme


Readme.md
```diff
@@ -90,6 +90,13 @@ data  = torch.zeros(2, 5, 2)
 output, new_state = layer.forward(data, state) 
 ```
 
+### Example on using the library: Low-level spike raster plots
```
cpehle: I think it might be better not to add this to the Readme in this  form. It is already rather long and the spike rasterplot is not that informative as is.

notebooks/norse_tutorial.ipynb
```diff
@@ -1,868 +1,1010 @@
...
-        }
-      },
-      "source": [
-        "import torchvision\n",
-        "\n",
-        "BATCH_SIZE = 32\n",
-        "\n",
-        "transform = torchvision.transforms.Compose(\n",
-        "    [\n",
-        "        torchvision.transforms.ToTensor(),\
```
cpehle: While it is definitely good that you ran the notebook to conclusion. The fact that we only reach 98% is actually a bug, because in internal code I easily reached >99%. I'm looking into a fix for that now.

Readme.md
Jegp: Sounds reasonable. How about moving it to the docs instead?

notebooks/norse_tutorial.ipynb
Jegp: Interesting! I didn't even look into the code.
I added a brief summary section in the end to conclude/summarise/perspectivise. Any new results that are competing with state-of-the-art would be great to include there!

Readme.md
Jegp: I removed it from the main docs and moved it to the tutorial. We still have the spiking notebook to refer to as well.

notebooks/norse_tutorial.ipynb
cpehle: I agree, that we can address this in a future patch. I've now played a bit with the network architecture, maybe that was the cause of the discrepancy. I will upload a new version of the tutorial notebook, once I've figured it out.

PR #20, opened 2020-03-05
Feature linting

Jegp:


cpehle:
I think in terms of whitespace changes, we should go with whatever black suggests and disable pylint complaints about it.

Jegp:
There's still a complexity check that fails for the cifar and cartpole task. I didn't pick it up because it'll likely be refactored when we start adding more examples anyway.

PR #43, opened 2020-05-16
Add torchaudio based speech commands dataset

cpehle:
This adds a wrapper for the SPEECHCOMMANDS dataset in torchaudio, which we can use for implementing an example task that requires temporal processing.


Jegp:
Should we have some form of regression test? Just to ensure that the API works going forward. 
Otherwise it looks good.

Issue #47, opened 2020-05-17
Inconsistent sparse latency encoder


Jegp:
There's a bug in the sparse latency encoder. Because `torch.topk` is inconsistent, the spike indices are not always the first.

PR #48, opened 2020-05-17
Bugfix latency encode

Jegp:
Fixer #47


norse/torch/module/test/test_encode.py
```diff
@@ -35,8 +35,15 @@ def spike_latency_encode_test():
...
-    for i, _ in enumerate(expected):
-        expected[i] = np.array([[0, 1, 1], [1, 0, 0]])
+    expected[0] = np.array([[0, 1, 1], [1, 1, 1]])
     np.testing.assert_equal(actual.numpy(), expected)
+
+
+def spike_latency_encode_chain_test():
+    data = torch.randn(7, 5) + 10
+    encoder = torch.nn.Sequential(
+        encode.ConstantCurrentLIFEncoder(2), encode.SpikeLatencyEncoder()
```
cpehle, line 47: I would test it out for a longer sequence and then assert that it only produces number of neurons spikes.

norse/torch/functional/encode.py
```diff
@@ -202,42 +202,10 @@ def spike_latency_encode(input_spikes: torch.Tensor) -> torch.Tensor:
...
-
-    spike_data = torch.take(input_spikes, spike_indices)
-
-    return torch.sparse_coo_tensor(
-        index_coordinates, spike_data, size=input_spikes.size()
-    )
+    mask = torch.zeros(input_spikes.size()[1:], device=input_spikes.device).byte()
+    spikes = input_spikes.clone()
+    zero_spikes = torch.zeros_like(spikes[0])
+    for index, spike_list in enumerate(spikes):
```
cpehle, line 208: It is probably much faster to do a range loop over the number of time steps.

norse/torch/functional/encode.py
```diff
@@ -202,42 +202,10 @@ def spike_latency_encode(input_spikes: torch.Tensor) -> torch.Tensor:
...
-        index_list.append(torch.take(dim_indices, spike_indices))
-    index_coordinates = torch.stack(index_list)
-
-    spike_data = torch.take(input_spikes, spike_indices)
-
-    return torch.sparse_coo_tensor(
-        index_coordinates, spike_data, size=input_spikes.size()
-    )
+    mask = torch.zeros(input_spikes.size()[1:], device=input_spikes.device).byte()
+    spikes = input_spikes.clone()
```
cpehle, line 206: If you want to return a copy, recommended practice is to use a list and then torch.stack at the end.

norse/torch/functional/encode.py
Jegp: Wow, yeah. Good catch. This is basically suicide

Jegp: That makes sense. We could rotate the spikes, go through the timesteps and then reshape. But would that be significantly more performant?

cpehle: Well I would do something like
```python
seq_length = spikes.shape[0]

for ts in range(seq_length):
    ...
```

Jegp: But isn't that identical to what I'm doing now? I'm basically iterating on the outer dimension?

norse/torch/functional/encode.py
Jegp: Done. Thanks!

norse/torch/functional/encode.py
cpehle: Well I'm pretty sure that this version would be optimised by the jit compiler (https://pytorch.org/blog/optimizing-cuda-rnn-with-torchscript/), in the other case I'm not sure if it would.

Jegp: Right. The link just says they "unroll loops". I'm indifferent here. This seems like such an obvious case to JIT optimise. I'll change the code just to be sure, but this would be interesting to explore.

PR #55, opened 2020-05-28
Fix notebooks

Jegp:
Ran the tutorial with the new encoders, corrected the other notebook slightly, and gave it a more meaningful name.


norse/benchmark/norse_lif.py
```diff
@@ -0,0 +1,107 @@
...
+flags.DEFINE_integer("stop", 10000, "Number of inputs to sweep to")
+flags.DEFINE_integer("sequence_length", 1000, "Number of timesteps to simulate")
+flags.DEFINE_float("dt", 0.001, "Simulation timestep")
+flags.DEFINE_string("device", "cpu", "Device to use [cpu, cuda]")
+
+
+import csv
+import torch
+import time
+import numpy as np
```
cpehle, line 32: I think those aren't used

PR #59, opened 2020-06-08
Add biologically plausible default parameters

cpehle:
This patch adds biologically plausible default parameters for two of the neuron models. Default parameters were taken from 
https://github.com/NeuralEnsemble/PyNN/blob/master/pyNN/standardmodels/cells.py
We should potentially cross reference with the literature for parameter ranges and implement a sampling scheme.


Jegp:
Good and elegant solution by just having a constant. 

My only comment would be to maybe add the PyNN reference in a comment.

Issue #63, opened 2020-06-09
Remove tensorboard as a dependency of the MNIST example task


cpehle:
This reduces the number of dependencies of the example considerably.

Issue #62, opened 2020-06-09
Question: is grpcio necessary?


dimaqq:
Pardon a silly question: what is grpcio needed for, is it essential?

The reason is that installing `norse` on py3.9 takes really long, as I guess they don't publish compiled wheels, probably because pypa didn't publish a manylinux container, because there's some confusion about ABI that could still possibly change...

Jegp:
Not a silly question at all. We're actually actively working on trying to trim the dependency tree. The `grpcio` is a dependency from the tensorboard package which, strictly speaking, doesn't have to be a part of Norse. So, thanks for pointing it out. I'll work on removing the dependency on Tensorboard.

Can I ask what environment you're in? Are you installing this from a container or a cutting-edge distro? I guess Python3.9 is only just emerging from staging.

cpehle:
The only hard dependency of the library itself is PyTorch. How to install PyTorch varies from your specific environment and is explained [here](https://pytorch.org/get-started/locally/). If you use Anaconda one way to setup a minimal environment is
```
conda create -n norse_test python=3.8 # or 3.9, but that is not available to me
conda activate norse_test
conda install pytorch torchvision -c pytorch # you should use whatever is recommended in the pytorch documentation here
conda install absl-py # command line parsing / logging used in example tasks
```
Right now the MNIST example task also uses tensorboard for logging, but I just created a pull request that makes that dependency optional. Once that is landed you should be able to run 
```
python -m norse.task.mnist
```
after completing the steps above.

dimaqq:
> Can I ask what environment you're in? Are you installing this from a container or a cutting-edge distro? I guess Python3.9 is only just emerging from staging.

I was trying to set it up on macOS, Python from python.org in a venv; with other bits (like hdf5) installed with `brew`. I didn't get very far 😅

Jegp:
Thanks for the reply @dimaqq. It's helpful for us to understand which environments Norse are being used in so we can work to accommodate them. We'll look into testing Python3.9 support. But, yeah. Unfortunately we can't help you with the rest :wink: Best of luck!

And please let us know if you have any other comments/critique for us! :slightly_smiling_face:

PR #64, opened 2020-06-09
Remove tensorboard + fix MNIST example task

cpehle:
This fixes the MNIST example task (got broken during refactoring) and makes tensorboard an optional dependency (this addresses #63, #62).


Jegp:
Neat-o!

If this is the only dependency on Tensorboard, it might make sense to recreate `requirements.txt` and `requirements-dev.txt`

PR #65, opened 2020-06-09
Feature remove grpcio dependency

### LINKED ISSUE #62: Question: is grpcio necessary?

Jegp:
Fixes #62


Jegp:
Alright, I've removed the task and added some errors I caught by running a minimal version of all the tasks. They should all be working. Could you verify @cpehle?

Jegp:
Codacy is complaining because the `main` function in one of the tasks is too complex. Don't think that's something we should tamper with now...

PR #69, opened 2020-06-12
Fixed typos and clarified some sentences in README

Jegp:


cpehle:
I made some further changes. I think it is important to be really conservative with regard to performance statements. Especially because some of the existing simulators like Nest easily scale to full HPC installations.

Jegp:
Completely agreed. It's definitely a good idea to specify what we *can* and *want* to be doing. Thanks for the edit :-)

PR #82, opened 2020-06-25
Speech commands task

cpehle:
This is a draft demonstrating audio processing on the google speech commands task.
There is still work to do regarding accuracy (the LIF model reaches ~64% after one epoch
but doesn't improve from there, which is suspicious) and even the processing of the dataset.

### DIFF
--- a/docs/source/experiments.rst
+++ b/docs/source/experiments.rst
@@ -89,3 +89,14 @@ The task can be run without any additional dependencies like so:
 .. code:: bash
 
     python3 -m norse.task.mnist
+
+Speech Command recognition task
+===============================
+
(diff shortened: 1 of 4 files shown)


Jegp:
Looks neat for now. Later it would be nice to have parameters, and maybe even semi-standardise them between tasks?
Nice idea for a task! :-)

Jegp:
Excellent work. Before merging, I think we should document how to run the task in both the [task readme](https://github.com/norse/norse/blob/master/norse/task/README.md) and [in the documentation](https://norse.github.io/norse/experiments.html).

It would also be nice with graphs, but we can always add that later :-)

PR #84, opened 2020-06-29
Added state capture and key-value arguments for LSNN layer

Jegp:


norse/torch/module/lsnn.py
```diff
@@ -107,25 +107,47 @@ class LSNNLayer(torch.nn.Module):
...
+            a) a single LSNNState, if return_states is set to False (or not set) or
+            b) a list of LSNNStates, if return_states is set to True
+        """
         inputs = input_tensor.unbind(0)
+        states = []
         outputs = []  # torch.jit.annotate(List[torch.Tensor], [])
         for i in range(len(inputs)):
             out, state = self.cell(inputs[i], state)
             outputs += [out]
+            if return_states:
```
cpehle, line 147: I think it would be preferable to collect the state of the voltages, currents and b separately and return them as stacked tensors in an LSNNState.

Jegp: I like the idea, but I can see how it's slightly confusing to get a tensor out with an implicit map of v/i/bs. How about reusing the state object then and returning the voltages, currents and biases in the same objects, but stacked? Then the forward pass would always return a `LSNNState` object. We could call the parameter something like `keep_state=False` where each component would be of shape (batches, n) if set to `False`, but shape (timesteps, batches, n) if set to `True`.

Jegp: Alright, this is actually quite annoying because it means we can't use the stacked LSNNState in forward computations. Unless we then later filter out the data. But I don't really see any neat solutions to this. We can either:

* Return a single state / list like before
* Return a single state, but stacked. That will lead to code like this:
 
```python
for ts in ...:
    _, s = lsnn(...)
    s = LSNNState(s.v[-1], s.i[-1], s.b[-1], s.v[-1])
```
* Return a single state with an added triplet if the boolean is set
* ?

PR #91, opened 2020-07-20
C++ implementation

cpehle:
This pull requests tracks the progress of the C++ implementation of norse primitives.

### DIFF
--- a/norse/torch/functional/lif.py
+++ b/norse/torch/functional/lif.py
@@ -1,8 +1,8 @@
-from typing import NamedTuple, Tuple
+from typing import NamedTuple, Optional, Tuple
 
 import torch
 import torch.jit
-
+import norse
 from norse.torch.functional.threshold import threshold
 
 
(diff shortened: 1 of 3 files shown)


Jegp:
Can we delete this PR?

cpehle:
I will update this branch with the remaining C++ commits, committing directly to master wasn't really intended.. :).

Jegp:
Looks good! Isn't this supposed to replace the lif_step function? Such that we remove the lif and rely on this instead?

Can I also suggest adding a simple regression test to verify that it'll work in the future?
Adding graphs would also be super cool, but not necessary for now I guess :-)

cpehle:
> Looks good! Isn't this supposed to replace the lif_step function? Such that we remove the lif and rely on this instead?

For now I think it would be best to leave both, to catch any regressions. Also in its current form it unfortunately can't replace it one to one, because it uses tuples instead of named tuples and apparently it isn't possible to return nested tuples easily.

> Can I also suggest adding a simple regression test to verify that it'll work in the future?

Sure, I will do that before this will be merged. Right now the build is cmake based so it takes a bit of work to integrate it with the CI.

Jegp:
> For now I think it would be best to leave both, to catch any regressions. Also in its current form it unfortunately can't replace it one to one, because it uses tuples instead of named tuples and apparently it isn't possible to return nested tuples easily.
Ok? Would it be a solution to solely use named tuples in the Python code? That shouldn't break native pattern matching patterns (`_, _ = ...`).

> Sure, I will do that before this will be merged. Right now the build is cmake based so it takes a bit of work to integrate it with the CI.
Excellent! :-)

Jegp:
I implemented the `lif_step` function and added a JIT optimized version as well. A quick benchmark tell me that we get a ~30% speedup for LIF. Which is nice.
I left out the `lif_integrate` for now. But this should get us started towards more performant neuron primitives. Next steps could be to 1) implement this for the functional `lift` and then 2) apply this to the layer abstractions as well.

norse/torch/functional/lif.py
```diff
@@ -111,51 +167,33 @@ def lif_step(
...
+        jit_params = LIFParametersJIT(
+            tau_syn_inv=p.tau_syn_inv,
+            tau_mem_inv=p.tau_mem_inv,
+            v_leak=p.v_leak,
+            v_th=p.v_th,
+            v_reset=p.v_reset,
+            method=p.method,
+            alpha=torch.as_tensor(p.alpha),
+        )
+        return _lif_step_jit(
```
Jegp, line 194: The only argument for having a JIT version next to the CPP version is that it *should* run when there are no binaries available (for instance on Windows?). But I haven't tested this.

Jegp (approved):

cpehle:
> I implemented the `lif_step` function and added a JIT optimized version as well. A quick benchmark tell me that we get a ~30% speedup for LIF. Which is nice.
> I left out the `lif_integrate` for now. But this should get us started towards more performant neuron primitives. Next steps could be to 1) implement this for the functional `lift` and then 2) apply this to the layer abstractions as well.

I think by far the biggest benefit will happen once we use a more clever event handling routine. The matrix multiplication we use right now is convenient but has much worse complexity than the translation by the weight vector that actually typically happens on spikes.

Jegp:
> I think by far the biggest benefit will happen once we use a more clever event handling routine. The matrix multiplication we use right now is convenient but has much worse complexity than the translation by the weight vector that actually typically happens on spikes.

Indeed. We might be able to learn something from the GeNN guys actually.
Thanks for reviewing!

PR #96, opened 2020-07-28
IBM DVS gesture dataset

cpehle:
This pull request implements a basic version of a torch wrapper for the IBM gesture dataset (http://www.research.ibm.com/dvsgesture/). What is missing is an automatic way of downloading the dataset
and a way to split it into train and validation datasets. We should probably follow the original paper in how to do that.

### DIFF
--- a/norse/dataset/dvs_gesture.py
+++ b/norse/dataset/dvs_gesture.py
@@ -1,0 +1,25 @@
+import torch
+import aedat
+import glob
+import os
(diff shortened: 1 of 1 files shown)


Jegp:
For the DVS dataset, doesn't it make sense to only do the processing when then `__getitem__` is called? Otherwise it'll take up 20> Gb of RAM upfront...

PR #99, opened 2020-08-05
Added exponential and adaptive LIF models

### LINKED ISSUE #60: Exponential Integrate and Fire neuron model
This is a standard pyNN model, that should be relatively easy to implement. Reference is

Brette R and Gerstner W (2005) Adaptive Exponential Integrate-and-Fire Model
    as an Effective Description of Neuronal Activity. J Neurophysiol 94:3637-3642.

See the corresponding pyNN  [implementation](https://github.com/NeuralEnsemble/PyNN/blob/d8056fa956998b031a1c3689a528473ed2bc0265/pyNN/standardmodels/cells.py#L416).

Jegp:
Fixes #60


Jegp:
Agree on all of these and we should document it somehow to avoid loosing these thoughts. Should we perhaps start some kind of vision/medium-term task document? We could use issues, but it seems to me these are less actionable while still being important objectives we want to work towards.

Issue #98, opened 2020-08-05
Does norse support GPU acceleration?


kongdelei:
Does norse support GPU acceleration?

Jegp:
Yes :-) Norse is built on [PyTorch](https://pytorch.org/). There's a nice description on how this works in PyTorch here: https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html?highlight=gpu

For Norse, you can basically take all the neuron models or layers and add a `.to('cuda')`. Here's a small example of a recurrent layer
```python
from norse.torch.module.lif import LIFCell
LIFCell(10, 20).to('cuda')
```

For more elaborate example with more layers, see for instance the [MNIST example task](https://github.com/norse/norse/blob/master/norse/task/mnist.py#L278).

Did that answer your question?

Jegp:
I added an article on the matter here: https://norse.github.io/norse/hardware.html

Closing for now. Feel free to reopen or ask another question :-)

PR #100, opened 2020-08-05
Added page on hardware and improved doc styling

Jegp:
I was annoyed with some of the documentation styling, so I made distinctions between functions/classes clearer, removed some annoying numbering, and made the page wider.

Also added a page on hardware acceleration as a reply to #98


Jegp:
Completely agree. I'll open an issue.

PR #104, opened 2020-08-18
Feature regularization

Jegp:
Implemented regularization by adding stateless functional accumulators as well as a stateful `RegularizationCell` module.
I chose not to implement a full `RegularizationState` NamedTuple just to avoid having to wrap the code in more types. Currently it should be straight-forward to both use and extend:

        >>> import torch
        >>> from norse.torch.module import lif, regularization
        >>> cell = lif.LIFCell(2, 4) # 2 -> 4
        >>> r = regularization.RegularizationCell() # Defaults to spike counting
        >>> data = torch.ones(5, 2)  # Batch size of 5
        >>> s = cell.initial_state(5, device='cpu')
        >>> z, s = cell(data, s)
        >>> z, rs = r(z, s)
        >>> print(rs) # 0 spikes yet

### DIFF
--- a/norse/__init__.py
+++ b/norse/__init__.py
@@ -1,7 +1,7 @@
 """Norse is a library for doing deep learning with spiking neural networks.
 """
 
-from . import benchmark, task
+from . import benchmark, dataset, task
 
 from .torch import functional
 from .torch.models import conv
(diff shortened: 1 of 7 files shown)


cpehle (approved):
This is a good start! I think we should consider adding torch.nn.Functions as well, which in their backward pass can then inject some temporally local regularisation term. But this can be another pull request.

PR #105, opened 2020-08-22
Online STDP learning rule

Huizerd:
Created a draft for online STDP, both multiplicative (soft and hard) and additive. Works with `nn.Linear`, `nn.Conv2d`, and `LIFCell`, as it just gives you the weight changes (which you can then apply yourself). Any comments about the implementation are very welcome, I just took some inspiration from the `correlation_sensor.py`, but maybe there are better ways of implementing this.

### DIFF
--- a/norse/torch/functional/stdp.py
+++ b/norse/torch/functional/stdp.py
@@ -1,60 +1,358 @@
-from typing import Callable, NamedTuple
+import torch
+
+from typing import Tuple, NamedTuple, Union
+
+from .heaviside import heaviside
+
+
+class STDPParameters(NamedTuple):
+    """Parameters of spike-timing-dependent plasticity (STDP)
+
+    Parameters:
+        a_pre (float): contribution of presynaptic spikes to trace
+        a_post (float): contribution of postsynaptic spikes to trace
+        tau_pre_inv (float): inverse decay time constant of presynaptic spike trace in 1/ms
+        tau_post_inv (float): inverse decay time constant of postsynaptic spike trace in 1/ms
+        w_min (float): lower bound on synaptic weight (should be < w_max)
+        w_max (float): upper bound on synaptic weight (should be > w_min)
+        eta_plus (float): contribution of pre-post activity to synaptic weight (should be >= 0)
+        eta_minus (float): contribution of post-pre activity to synaptic weight (should be >= 0)
+    """
+    a_pre: torch.Tensor = torch.as_tensor(1.0)
+    a_post: torch.Tensor = torch.as_tensor(1.0)
+    tau_pre_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    tau_post_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    w_min: float = 0.0
+    w_max: float = 1.0
+    eta_plus: float = 1.0
+    eta_minus: float = 1.0
+
+
+# TODO: can be merged with above, but redundant passing around of values
+class STDPConvParameters(NamedTuple):
+    """Parameters of spike-timing-dependent plasticity (STDP). Same as STDPParameters except
+    additional information on convolution (e.g., stride, padding, dilation)
+
+    Parameters:
+        a_pre (float): contribution of presynaptic spikes to trace
+        a_post (float): contribution of postsynaptic spikes to trace
+        tau_pre_inv (float): inverse decay time constant of presynaptic spike trace in 1/ms
+        tau_post_inv (float): inverse decay time constant of postsynaptic spike trace in 1/ms
+        w_min (float): lower bound on synaptic weight (should be < w_max)
+        w_max (float): upper bound on synaptic weight (should be > w_min)
+        eta_plus (float): contribution of pre-post activity to synaptic weight (should be >= 0)
+        eta_minus (float): contribution of post-pre activity to synaptic weight (should be >= 0)
+        stride (int or tuple): convolutional stride, defaults to 1 like nn.Conv2d
+        padding (int or tuple): convolutional padding, defaults to 0 like nn.Conv2d
+        dilation (int or tuple): convolutional dilation, defaults to 1 like nn.Conv2d
+    """
+    a_pre: torch.Tensor = torch.as_tensor(1.0)
+    a_post: torch.Tensor = torch.as_tensor(1.0)
+    tau_pre_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    tau_post_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    w_min: float = 0.0
+    w_max: float = 1.0
+    eta_plus: float = 1.0
+    eta_minus: float = 1.0
+    stride: Union[int, Tuple] = 1
+    padding: Union[int, Tuple] = 0
+    dilation: Union[int, Tuple] = 1
+
+
 
 
 class STDPState(NamedTuple):
-    """State of an event driven STDP.
+    """State of spike-timing-dependent plasticity (STDP)
 
     Parameters:
-        a_pre (torch.Tensor): presynaptic STDP state.
-        a_post (torch.Tensor): postsynaptic STDP state.
+        x (torch.Tensor): presynaptic spike trace
+        y (torch.Tensor): postsynaptic spike trace
     """
-
-    a_pre: torch.Tensor
-    a_post: torch.Tensor
-
-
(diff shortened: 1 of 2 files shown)


norse/torch/functional/stdp.py
```diff
@@ -1,60 +1,407 @@
...
+    a_post: torch.Tensor = torch.as_tensor(1.0)
+    tau_pre_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    tau_post_inv: torch.Tensor = torch.as_tensor(1.0 / 100e-3)
+    w_min: float = 0.0
+    w_max: float = 1.0
+    eta_plus: float = 1.0
+    eta_minus: float = 1.0
+
+
+# TODO: can be merged with above, but redundant passing around of values
```
Jegp, line 32: I think splitting these up is a good idea to ensure disambiguation

norse/torch/functional/stdp.py
```diff
@@ -1,60 +1,407 @@
...
+
+    dy = dt * p.tau_post_inv * (-state.y + p.a_post * z_post)
+    y_decayed = state.y + dy
+
+    # w has shape (post, pre), so spikes/trace for postsynaptic neuron have to be transposed
+    # Pre- and postsynaptic neurons can be any 2-4D (BCHW) shape
+    dw_plus = _A_plus_soft(w, p.w_max, p.eta_plus) * x_decayed * z_post[..., None]
+    dw_minus = _A_minus_soft(w, p.w_min, p.eta_minus) * y_decayed[..., None] * z_pre
+
+    # TODO: what to do with batch dimension in an online rule? Sum? Nothing?
```
Jegp, line 153: Since you're not explicitly changing the weights, I think this is a sound approach. If batch normalisation or similar is required, it's entirely possible to implement that afterwards. 
It would also be a possibility to just expose al batches as-is. Would that be desirable in any situation?

Jegp (commented):
Excellent work @Huizerd! I like how you've kept the immutability and freedom of implementing the weight updates.
I left some comments inline. Apart from that I have three thoughts

* Brilliant you are testing for functional code. Would it make sense to add a few assertions just to make sure that some basic principles around the weight updates hold? I know our code use some of the same logic, but we're trying to change that so we have more useful regression tests.
* It would also be great with a module implementation of a (stateful) STDP layer. We can also work on that later if you prefer.
* I noticed you use a fixed heaviside function (from `heaviside.py`). Would it make sense to parameterise that, as done (quite ugly, sorry @cpehle :-) ) in `threshold.py`?

norse/torch/functional/stdp.py
Huizerd: I'm not sure whether that would be desirable.. However, there might be multiple options for combining batches (mean, sum), so it would be most flexible to leave this choice to the user. Both BindsNET and PySNN do some sort of aggregation though I think

norse/torch/functional/stdp.py
Huizerd: Ok!

Huizerd:
Thanks!

>     * Brilliant you are testing for functional code. Would it make sense to add a few assertions just to make sure that some basic principles around the weight updates hold? I know our code use some of the same logic, but we're trying to change that so we have more useful regression tests.

Yes, this is something I am going to add for sure. (I actually already converted some of the existing tests for LIF neurons to check for the actual voltages coming out, but haven't pushed those yet :))

>     * It would also be great with a module implementation of a (stateful) STDP layer. We can also work on that later if you prefer.

That makes sense to have, yes. I'll look at that for another PR!

>     * I noticed you use a fixed heaviside function (from `heaviside.py`). Would it make sense to parameterise that, as done (quite ugly, sorry @cpehle :-) ) in `threshold.py`?

You mean for allowing autograd gradients to flow through? Not sure that would add anything except extra computations, but I could be wrong :)

norse/torch/functional/stdp.py
Jegp: Hmm I always found the trade-off between freedom and usability quite hard. Do you have any opinions @cpehle?

Jegp:
> Yes, this is something I am going to add for sure. (I actually already converted some of the existing tests for LIF neurons to check for the actual voltages coming out, but haven't pushed those yet :))

Excellent! And much appreciated :-)

> You mean for allowing autograd gradients to flow through? Not sure that would add anything except extra computations, but I could be wrong :)

Am I understanding it correctly that you're right now clamping the `dw` to zero if the weights are outside `[w_min;wmax]` and to `eta` otherwise? That raises two questions for me. 1) What happens if the application of `eta` leads to a weight that exceeds `[w_min;wmax]`? And 2) would any other behaviour be desirable? Like, for instance, a smoothing towards the extremes.

Note that I haven't studied the STDP implementations closely and that I might be overthinking this. As you pointed out, the user is free to update the weights in any way (s)he wants.

norse/torch/functional/stdp.py
```diff
@@ -1,60 +1,407 @@
...
+
+    dy = dt * p.tau_post_inv * (-state.y + p.a_post * z_post)
+    y_decayed = state.y + dy
+
+    # w has shape (post, pre), so spikes/trace for postsynaptic neuron have to be transposed
+    # Pre- and postsynaptic neurons can be any 2-4D (BCHW) shape
+    dw_plus = p.eta_plus * x_decayed * z_post[..., None]
+    dw_minus = p.eta_minus * y_decayed[..., None] * z_pre
+
+    # TODO: what to do with batch dimension in an online rule? Sum? Nothing?
```
cpehle, line 226: See above

norse/torch/functional/stdp.py
cpehle: One option would be to pass in a function as an argument. It also should be possible to integrate weight changes over time (relative to the integration timestep for the neuron dynamics). So it might be necessary to carry though the dw_+, dw_- changes over multiple time steps.

norse/torch/functional/stdp.py
```diff
@@ -1,60 +1,407 @@
...
+
+    Parameters:
+        w (torch.Tensor): synaptic weight
+        w_min (float): lower bound on synaptic weight (should be < w_max)
+        eta_minus (float): contribution of post-pre activity to synaptic weight (should be >= 0)
+    """
+    return heaviside(w - w_min) * eta_minus
+
+
+def linear_soft_multiplicative_stdp_step(
```
cpehle, line 122: I think this function and the function below might benefit from being a partial specialisations of a general function. We don't need to do it in this pull request. But it should happen soon.

norse/torch/functional/stdp.py
cpehle: I think it might be possible to merge the two if the STDP computation below is carried out on tensors or alternatively the update rule is a torch.nn.Module that carries these parameters.

cpehle (approved):
Thanks for the great contribution 👍. I've left some comments, but those could also be addressed in a further pull request.

norse/torch/functional/stdp.py
Jegp: Good point. Would you mind creating an issue just to avoid loosing this gem of wisdom? :-)

norse/torch/functional/stdp.py
Jegp: It seems like this isn't exactly a red lamp in the sense that it'll become an immediate problem. Would you mind if I resolved this conversation for now and then we can always add the feature later, if needed?

Huizerd:
Thanks for the comments! I'll work on them over the coming days.

> Am I understanding it correctly that you're right now clamping the dw to zero if the weights are outside [w_min;wmax] and to eta otherwise? That raises two questions for me. 1) What happens if the application of eta leads to a weight that exceeds [w_min;wmax]? And 2) would any other behaviour be desirable? Like, for instance, a smoothing towards the extremes.

The `hard_multiplicative` variants use a Heaviside function, but the `soft_multiplicative` use a linearly decaying function. I just mostly followed what was already in the `stdp.py` file and [scholarpedia](http://www.scholarpedia.org/article/Spike-timing_dependent_plasticity), but other variants are possible of course.

norse/torch/functional/stdp.py
cpehle: Good point I've done so here: #107.

Jegp (approved):

Jegp:
Thanks again for all your work and for interacting with us @Huizerd. To me, all the changes looks sound and I think it's a solid beginning. We can always look into future nice-to-haves in another PR.

PR #106, opened 2020-08-25
Remove redundant initial_state by inferring device and dtype

### LINKED ISSUE #101: Remove device parameters from neuron models
We are currently using `device` parameters for some neuron models (like ConvNet). Instead, it should be applied on the top level and the models themselves should inherit device parameters from passed data.

- [x] Remove device parameters from neuron models
- [x] Update code examples
- [x] Update tasks
- [x] Update hardware documentation page

Jegp:
I solved this by removing all the device parameters from the neuron models and adding a `.to` method for all the `NamedTuple`s encoding state. 
It's slightly hacky but I think it's a solid solution, compared to the alternative of either a) inferring device on the fly (from where?) and b) mapping all tensors to stateful buffers.

Fixes #101

### DIFF
--- a/norse/benchmark/main.py
+++ b/norse/benchmark/main.py
@@ -58,7 +58,9 @@ def benchmark(
                 logging.error(message)
 
             data = BenchmarkData(
-                config=config, durations=np.array(durations), parameters=parameters,
+                config=config,
+                durations=np.array(durations),
+                parameters=parameters,
             )
             result = collector(data)
 
--- a/norse/examples/plot_mnist.py
+++ b/norse/examples/plot_mnist.py
@@ -25,7 +25,6 @@ from norse.torch.module.lif import LIFFeedForwardCell
 class Net(torch.nn.Module):
     def __init__(
         self,
-        device="cpu",
         num_channels=1,
         feature_size=32,
         model="super",
@@ -49,30 +48,24 @@ class Net(torch.nn.Module):
             (1024,), p=LIFParameters(method=model, alpha=100.0)
         )
         self.out = LICell(1024, 10)
-        self.device = device
         self.dtype = dtype
+        # One would normally also define the device here
+        # However, Norse has been built to infer the device type from the input data
+        # It is still possible to enforce the device type on initialisation
+        # More details are available in our documentation:
+        #   https://norse.github.io/norse/hardware.html
 
     def forward(self, x):
         seq_length = x.shape[0]
         seq_batch_size = x.shape[1]
 
         # specify the initial states
-        s0 = self.lif0.initial_state(
-            seq_batch_size, device=self.device, dtype=self.dtype
-        )
-        s1 = self.lif1.initial_state(
-            seq_batch_size, device=self.device, dtype=self.dtype
-        )
-        s2 = self.lif2.initial_state(
-            seq_batch_size, device=self.device, dtype=self.dtype
-        )
-        so = self.out.initial_state(
-            seq_batch_size, device=self.device, dtype=self.dtype
-        )
+        s0 = None
+        s1 = None
+        s2 = None
+        so = None
 
-        voltages = torch.zeros(
-            seq_length, seq_batch_size, 10, device=self.device, dtype=self.dtype
-        )
+        voltages = torch.zeros(seq_length, seq_batch_size, 10, dtype=self.dtype)
 
         for ts in range(seq_length):
             z = self.conv1(x[ts, :])
--- a/norse/task/cartpole.py
+++ b/norse/task/cartpole.py
@@ -54,13 +54,12 @@ class ANNPolicy(torch.nn.Module):
 
 
 class Policy(torch.nn.Module):
-    def __init__(self, device="cpu"):
+    def __init__(self):
         super(Policy, self).__init__()
         self.state_dim = 4
         self.input_features = 16
         self.hidden_features = 128
         self.output_features = 2
-        self.device = device
         self.constant_current_encoder = ConstantCurrentLIFEncoder(40)
         self.lif = LIFCell(
             2 * self.state_dim,
@@ -75,22 +74,17 @@ class Policy(torch.nn.Module):
 
     def forward(self, x):
         scale = 50
-        x = x.to(self.device)
         x_pos = self.constant_current_encoder(torch.nn.functional.relu(scale * x))
         x_neg = self.constant_current_encoder(torch.nn.functional.relu(-scale * x))
         x = torch.cat([x_pos, x_neg], dim=2)
 
         seq_length, batch_size, _ = x.shape
 
-        # state for hidden layer
-        s1 = self.lif.initial_state(batch_size, device=self.device)
-        # state for output layer
-        so = self.readout.initial_state(batch_size, device=self.device)
-
         voltages = torch.zeros(
-            seq_length, batch_size, self.output_features, device=self.device
+            seq_length, batch_size, self.output_features, device=x.device
         )
 
+        s1 = so = None
         # sequential integration loop
         for ts in range(seq_length):
             z1, s1 = self.lif(x[ts, :, :], s1)
@@ -104,13 +98,12 @@ class Policy(torch.nn.Module):
 
 
 class LSNNPolicy(torch.nn.Module):
-    def __init__(self, device="cpu", model="super"):
+    def __init__(self, model="super"):
         super(LSNNPolicy, self).__init__()
         self.state_dim = 4
         self.input_features = 16
         self.hidden_features = 128
         self.output_features = 2
-        self.device = device
         # self.affine1 = torch.nn.Linear(self.state_dim, self.input_features)
         self.constant_current_encoder = ConstantCurrentLIFEncoder(40)
         self.lif_layer = LSNNCell(
@@ -133,12 +126,12 @@ class LSNNPolicy(torch.nn.Module):
         seq_length, batch_size, _ = x.shape
 
         # state for hidden layer
-        s1 = self.lif_layer.initial_state(batch_size, device=self.device)
+        s1 = None
         # state for output layer
-        so = self.readout.initial_state(batch_size, device=self.device)
+        so = None
 
         voltages = torch.zeros(
-            seq_length, batch_size, self.output_features, device=self.device
+            seq_length, batch_size, self.output_features, device=x.device
         )
 
         # sequential integration loop
@@ -206,7 +199,7 @@ def main(args):
     elif FLAGS.policy == "snn":
         policy = Policy()
     elif FLAGS.policy == "lsnn":
-        policy = LSNNPolicy(device=FLAGS.device, model=FLAGS.model).to(FLAGS.device)
+        policy = LSNNPolicy(model=FLAGS.model).to(FLAGS.device)
     optimizer = torch.optim.Adam(policy.parameters(), lr=FLAGS.learning_rate)
 
     running_rewards = []
--- a/norse/task/cifar10.py
+++ b/norse/task/cifar10.py
@@ -114,22 +114,17 @@ def signed_poisson_train(images, seq_length, rel_fmax=0.2):
 
 
 class LIFConvNet(torch.nn.Module):
-    def __init__(self, num_channels, device="cpu"):
+    def __init__(self, num_channels):
         super(LIFConvNet, self).__init__()
 
         if FLAGS.net == "convnet":
             dtype = torch.float
-            self.rsnn = ConvNet(
-                device=device, num_channels=num_channels, feature_size=32, dtype=dtype
-            )
+            self.rsnn = ConvNet(num_channels=num_channels, feature_size=32, dtype=dtype)
         elif FLAGS.net == "convnet4":
-            self.rsnn = ConvNet4(
-                device=device, num_channels=num_channels, feature_size=32
-            )
-        self.device = device
+            self.rsnn = ConvNet4(num_channels=num_channels, feature_size=32)
 
     def forward(self, x):
-        voltages = self.rsnn(x.to(self.device).permute(1, 0, 2, 3, 4))
+        voltages = self.rsnn(x).permute(1, 0, 2)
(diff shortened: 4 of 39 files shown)


cpehle (commented):
Nice work :). I think this is a good approach, one thing I think you could investigate would be
- the torch.nn.Modules will have a "canonical" device on which their parameters are on. This doesn't work for feedforward cells though (https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently)
- A less severe question would be compatibility with torchjit, but this would be another issue
- Perhaps it would be possible to let NamedTupleTensor be a subclass of NamedTuple

Jegp:
>     * the torch.nn.Modules will have a "canonical" device on which their parameters are on. This doesn't work for feedforward cells though (https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently)

There are, but the tensors won't be moved if someone applies `.to` to the module/state. I also looked into the possibility of just extending NamedTuple, but that would force all state implementations to override the `__new__` method each time you have a new class. 

Both approaches would be cleaner, but wouldn't give us exactly what we want and wouldn't be as easy to use/implement. The whole point, to me, is to make it as simple as possible to use.

>     * A less severe question would be compatibility with torchjit, but this would be another issue

I don't know exactly what it takes to be compatible, but in principle we should be able to add whatever functionality we want to the `NamedTupleTensor` (meta)class. So we can make it behave in any way we want. I assume that should make it at least feasible to offer compatibility?

norse/benchmark/main.py
```diff
@@ -58,7 +58,9 @@ def benchmark(
                 logging.error(message)
 
             data = BenchmarkData(
-                config=config, durations=np.array(durations), parameters=parameters,
+                config=config,
```
cpehle, line 61: How did this whitespace change happen?

norse/examples/plot_mnist.py
```diff
@@ -49,30 +48,19 @@ def __init__(
...
         )
         self.out = LICell(1024, 10)
-        self.device = device
         self.dtype = dtype
 
     def forward(self, x):
         seq_length = x.shape[0]
         seq_batch_size = x.shape[1]
 
         # specify the initial states
```
cpehle, line 57: Since this is a tutorial I think there should be a comment here, explaining the initialisation

norse/task/cartpole.py
```diff
@@ -75,22 +74,15 @@ def __init__(self, device="cpu"):
...
 
-        # state for hidden layer
-        s1 = self.lif.initial_state(batch_size, device=self.device)
-        # state for output layer
-        so = self.readout.initial_state(batch_size, device=self.device)
-
-        voltages = torch.zeros(
-            seq_length, batch_size, self.output_features, device=self.device
-        )
+        voltages = torch.zeros(seq_length, batch_size, self.output_features)
```
cpehle, line 83: I think this still needs a device parameter, maybe use self.parameters[0].device.

norse/task/cartpole.py
```diff
@@ -133,13 +124,11 @@ def forward(self, x):
...
-        s1 = self.lif_layer.initial_state(batch_size, device=self.device)
+        s1 = None
         # state for output layer
-        so = self.readout.initial_state(batch_size, device=self.device)
+        so = None
 
-        voltages = torch.zeros(
-            seq_length, batch_size, self.output_features, device=self.device
-        )
+        voltages = torch.zeros(seq_length, batch_size, self.output_features)
```
cpehle, line 131: Same here.

norse/task/cifar10.py
```diff
@@ -114,22 +114,17 @@ def signed_poisson_train(images, seq_length, rel_fmax=0.2):
...
         elif FLAGS.net == "convnet4":
-            self.rsnn = ConvNet4(
-                device=device, num_channels=num_channels, feature_size=32
-            )
-        self.device = device
+            self.rsnn = ConvNet4(num_channels=num_channels, feature_size=32)
 
     def forward(self, x):
-        voltages = self.rsnn(x.to(self.device).permute(1, 0, 2, 3, 4))
+        voltages = self.rsnn(x).permute(1, 0, 2)
```
cpehle, line 127: Good catch

norse/task/correlation_experiment.py
```diff
@@ -91,9 +91,9 @@ def generate_random_data(
         for ts in range(seq_length):
             z1, s1 = lif_correlation(
                 x[ts, :, :],
-                s1,
                 input_weights=input_weights,
                 recurrent_weights=recurrent_weights,
+                state=s1,
```
cpehle, line 96: I'm a bit confused by this change

norse/torch/functional/threshold.py
```diff
@@ -90,7 +99,10 @@ def forward(ctx, x, alpha):
 
     @staticmethod
     def backward(ctx, dy):
-        x, alpha, = ctx.saved_tensors
+        (
```
cpehle, line 102: pretty strange whitespace change, see also above..

norse/torch/module/lif.py
```diff
@@ -79,16 +78,30 @@ def extra_repr(self):
...
     ) -> Tuple[torch.Tensor, LIFState]:
+        if state is None:
+            state = LIFState(
+                z=torch.zeros(
+                    input_tensor.shape[0],
+                    self.hidden_size,
+                    device=input_tensor.device,
+                    dtype=input_tensor.dtype,
+                ),
+                v=torch.zeros(
```
cpehle, line 92: self.p.v_leak * torch.ones(...)

cpehle: or a constant initialisation with self.p.v_leak

norse/torch/module/lif.py
```diff
@@ -180,13 +188,12 @@ def extra_repr(self):
...
-            i=torch.zeros(batch_size, *self.shape, device=device, dtype=dtype),
-        )
-
     def forward(
-        self, x: torch.Tensor, state: LIFFeedForwardState
+        self, x: torch.Tensor, state: Optional[LIFFeedForwardState] = None
     ) -> Tuple[torch.Tensor, LIFFeedForwardState]:
+        if state is None:
+            state = LIFFeedForwardState(
+                v=torch.zeros(x.shape[0], *self.shape, device=x.device, dtype=x.dtype),
```
cpehle, line 196: self.p.v_leak

norse/torch/module/lif_adex.py
```diff
@@ -80,17 +79,36 @@ def extra_repr(self):
...
     ) -> Tuple[torch.Tensor, LIFAdExState]:
+        if state is None:
+            state = LIFAdExState(
+                z=torch.zeros(
+                    input_tensor.shape[0],
+                    self.hidden_size,
+                    device=input_tensor.device,
+                    dtype=input_tensor.dtype,
+                ),
+                v=torch.zeros(
```
cpehle, line 93: self.p.v_leak

norse/torch/module/lif_adex.py
```diff
@@ -187,14 +200,28 @@ def extra_repr(self):
...
-            a=torch.zeros(batch_size, *self.shape, device=device, dtype=dtype),
-        )
-
     def forward(
-        self, x: torch.Tensor, state: LIFAdExFeedForwardState
+        self, x: torch.Tensor, state: Optional[LIFAdExFeedForwardState] = None
     ) -> Tuple[torch.Tensor, LIFAdExFeedForwardState]:
+        if state is None:
+            state = LIFAdExFeedForwardState(
+                v=torch.zeros(
```
cpehle, line 208: self.p.v_leak

norse/torch/module/lif_correlation.py
```diff
@@ -25,61 +25,78 @@ def __init__(
...
+            batch_size = input_tensor.shape[0]
+            state = LIFCorrelationState(
+                lif_state=LIFState(
+                    z=torch.zeros(
+                        batch_size,
+                        hidden_features,
+                        device=input_tensor.device,
+                        dtype=input_tensor.dtype,
+                    ),
+                    v=torch.zeros(
```
cpehle, line 47: self.p.v_leak

norse/torch/module/lif_ex.py
```diff
@@ -79,16 +78,29 @@ def extra_repr(self):
...
+        self, input_tensor: torch.Tensor, state: Optional[LIFExState] = None
     ) -> Tuple[torch.Tensor, LIFExState]:
+        if state is None:
+            state = LIFExState(
+                z=torch.zeros(
+                    (input_tensor.shape[0], self.hidden_size),
+                    dtype=input_tensor.dtype,
+                    device=input_tensor.device,
+                ),
+                v=torch.zeros(
```
cpehle, line 91: self.p.v_leak

norse/torch/module/lif_ex.py
```diff
@@ -184,13 +191,12 @@ def extra_repr(self):
...
-            i=torch.zeros(batch_size, *self.shape, device=device, dtype=dtype),
-        )
-
     def forward(
-        self, x: torch.Tensor, state: LIFExFeedForwardState
+        self, x: torch.Tensor, state: Optional[LIFExFeedForwardState] = None
     ) -> Tuple[torch.Tensor, LIFExFeedForwardState]:
+        if state is None:
+            state = LIFExFeedForwardState(
+                v=torch.zeros(x.shape[0], *self.shape, device=x.device, dtype=x.dtype),
```
cpehle, line 199: self.p.v_leak

norse/torch/module/lif_mc.py
```diff
@@ -65,18 +65,30 @@ def __init__(
...
     ) -> Tuple[torch.Tensor, LIFState]:
+        if state is None:
+            state = LIFState(
+                z=torch.zeros(
+                    input_tensor.shape[0],
+                    self.hidden_size,
+                    device=input_tensor.device,
+                    dtype=input_tensor,
+                ),
+                v=torch.zeros(
```
cpehle, line 79: self.p.v_leak

norse/torch/module/lif_mc_refrac.py
```diff
@@ -28,21 +28,38 @@ def __init__(
...
+        if state is None:
+            state = LIFRefracState(
+                lif=LIFState(
+                    z=torch.zeros(
+                        input_tensor.shape[0],
+                        self.hidden_size,
+                        device=input_tensor.device,
+                        dtype=input_tensor.dtype,
+                    ),
+                    v=torch.zeros(
```
cpehle, line 43: self.p.v_leak

norse/torch/module/lif_refrac.py
```diff
@@ -82,19 +81,38 @@ def __init__(
...
+        if state is None:
+            state = LIFRefracState(
+                LIFState(
+                    z=torch.zeros(
+                        input_tensor.shape[0],
+                        *self.shape,
+                        device=input_tensor.device,
+                        dtype=input_tensor.dtype
+                    ),
+                    v=torch.zeros(
```
cpehle, line 96: self.p.v_leak

norse/torch/module/lif_refrac.py
```diff
@@ -143,28 +161,46 @@ class LIFRefracFeedForwardCell(torch.nn.Module):
...
     def forward(
-        self, input_tensor: torch.Tensor, state: LIFRefracFeedForwardState
+        self,
+        input_tensor: torch.Tensor,
+        state: Optional[LIFRefracFeedForwardState] = None,
     ) -> Tuple[torch.Tensor, LIFRefracFeedForwardState]:
+        if state is None:
+            state = LIFRefracFeedForwardState(
+                LIFFeedForwardState(
+                    v=torch.zeros(
```
cpehle, line 186: self.p.v_leak

norse/torch/module/lsnn.py
```diff
@@ -70,18 +70,36 @@ def __init__(
...
     ) -> Tuple[torch.Tensor, LSNNState]:
+        if state is None:
+            state = LSNNState(
+                z=torch.zeros(
+                    input_tensor.shape[0],
+                    self.output_features,
+                    device=input_tensor.device,
+                    dtype=input_tensor.dtype,
+                ),
+                v=torch.zeros(
```
cpehle, line 84: self.p.v_leak

norse/torch/module/lsnn.py
```diff
@@ -181,17 +193,28 @@ def __init__(self, shape, p: LSNNParameters = LSNNParameters(), dt: float = 0.00
...
-            b=torch.zeros(batch_size, *self.shape, device=device, dtype=dtype),
-        )
-
     def forward(
-        self, input_tensor: torch.Tensor, state: LSNNFeedForwardState
+        self, input_tensor: torch.Tensor, state: Optional[LSNNFeedForwardState] = None
     ) -> Tuple[torch.Tensor, LSNNFeedForwardState]:
+        if state is None:
+            state = LSNNFeedForwardState(
+                v=torch.zeros(
```
cpehle, line 201: self.p.v_leak

cpehle (changes requested):
I found some minor things, we should consider running the benchmark to make sure that the performance doesn't suffer too much. Also there are some strange whitespace changes. Otherwise very nice work 👍 .

norse/benchmark/main.py
Jegp: Black got upgraded. I just added the new formatting because it's in our C

norse/task/cartpole.py
Jegp: I solved it with `x.device`

norse/task/correlation_experiment.py
Jegp: I changed the order of the parameter for `lif_correlation` because the `state` param now has a default value.

Jegp:
Thanks for a super thorough review @cpehle! Let me know if you have further comments :-)

Jegp:
I'm slightly annoyed with black having weird formatting rules. Do you know whether it's possible to configure?

cpehle (approved):
LGTM 👍.

PR #108, opened 2020-08-31
Use codecov.io to track code coverage

cpehle:


Jegp (approved):
Great changes :-)
Let's get it going

Issue #23, opened 2020-03-14
Use a static type checker like MyPy


Jegp:
We should type check the code (possibly as a GitHub workflow).
http://mypy-lang.org/ is one possible type checker

PR #111, opened 2020-09-03
Fixes necessary to make pytype happy

cpehle:
There are still some more files that don't typecheck using pytype.
It also needs to be integrated with GitHub workflow.


Jegp:
I guess this relates to #23?

cpehle:
> I guess this relates to #23?

Yes, although I think pytype seems like a better choice, because it also does type inference. This feature already caught some bugs.

Jegp (approved):

PR #117, opened 2020-09-27
Feature benchmark

Jegp:
Tests PyGeNN and implements C++ optimisation for superspike.


CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 project(norse)
 
 find_package(Torch REQUIRED)
-find_package (Python3 REQUIRED COMPONENTS Development)
+FIND_PACKAGE(PythonInterp 3 REQUIRED)
```
cpehle, line 5: These are both deprecated in favour of what was used before. Perhaps you are using an outdated version of CMake?

norse/__init__.py
```diff
@@ -3,6 +3,11 @@
 
 from . import benchmark, dataset, task
 
+import torch
+from pathlib import Path
+
+torch.ops.load_library(Path(__file__).resolve().parent.parent / "norse_op.so")
```
cpehle, line 9: Does this still work if we don't build the C++ library? Do we want to make it mandatory to build it?

norse/benchmark/README.md
```diff
@@ -4,16 +4,21 @@ We plan to extend it with more benchmarks in the immediate future.
...
+timesteps with a time-delta of 0.001 seconds and a batch-size of 32.
+Each line indicates the mean running time out of 20 repeated runs and the coloured
+area indicates where ~95% of the runtimes fell.
 
-The benchmark indicates that for single-layer LIF neurons, Norse outperforms
-both competitors.
+BindsNET ran out of memory in > 4500 neuron layers, indicated by the orange circle.
+
+The benchmark indicates that for single-layer LIF neurons of size < 4500, Norse outperforms both competitors.
+The overtaking of GeNN is likely due to the fact that they do not have matrix multiplications, which scales quadratically with size in Norse and BindsNET.
```
cpehle, line 20: have -> use?

norse/benchmark/bindsnet_lif.py
```diff
@@ -28,9 +28,12 @@ def lif_feed_forward_benchmark(parameters: BenchmarkParameters):
         .to(parameters.device)
         .float()
     )
+    input_spikes.requires_grad = False
 
     input_data = {"Input": input_spikes}
     network.to(parameters.device)
+    for param in network.parameters():
```
cpehle, line 35: Since we found that this actually makes performance worse, maybe omit?

cpehle (changes requested):
Great work overall. We should look into more systematic ways of integrating the C++ code next. But this is a good starting point.

norse/benchmark/bindsnet_lif.py
cpehle: But omit I mean don't set requires_grad to false.

CMakeLists.txt
Jegp: Likely. I tried again, and it seemed to work. Strange.

norse/__init__.py
Jegp: Great point. I found a solution where we basically set a flag to indicate whether the ops modules have been loaded. I tested whether it works both with and without. In the future, though, I think we should do something like the PyTorch guys, where modules are basically mirrored in Python and in compiled code.

norse/benchmark/bindsnet_lif.py
Jegp: Did we? I think we found that it improved the benchmark quite a bit... But I might be remembering wrongly?

cpehle: I thought the profiling traces for Norse showed that it made performance worse.

cpehle (approved):

norse/benchmark/bindsnet_lif.py
Jegp: Actually, yeah. It's pretty much the same:
image
At some point it gave us a 0.5x boost. But this might have been solved with the JIT

PR #120, opened 2020-10-01
Added minor tests to ensure lif state shape

Jegp:
Small tests to make sure the lif state is well formed and that the lif cells behaves well in sequence (not torch.nn.Sequential)


cpehle (approved):
LGTM

PR #122, opened 2020-10-02
Fixed LIFeedForward and documented LIFFeedForward

Jegp:
A fix for the leaky integrator feed forward layer

Also added some documentation and tests


cpehle (approved):
LGTM

PR #123, opened 2020-10-04
VGG models lifted to SNN models

cpehle:
This is the first of a series of commits that port torchvision models to have intermediate spiking layers (LIFFeedForwardLayers  for now). Some other alternatives could also be considered. We could introduce additional models, where the features are computed  by an ANN and convert only the classifier to an SNN. 

Loading weights from the ANN, pretrained on the ImageNet dataset  also works, but no tests on the classification accuracy has been done. It is unlikely to be high.


Jegp:
Beautiful! Nice and simple. Let's get it merged

Jegp (approved):

Issue #125, opened 2020-10-04
Fix gradient flow in LIF cells


Jegp:
Gradients are not being propagated properly with the new default states

Jegp:
Closed by #127

PR #127, opened 2020-10-04
Issue 125 gradient

Jegp:
Fixes issues with gradients not flowing correctly #125

### DIFF
--- a/norse/torch/module/lif_mc.py
+++ b/norse/torch/module/lif_mc.py
@@ -93,6 +93,7 @@ class LIFMCCell(torch.nn.Module):
                     dtype=input_tensor.dtype,
                 ),
             )
+            state.v.requires_grade = True
         return lif_mc_step(
             input_tensor,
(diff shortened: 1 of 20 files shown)


norse/torch/module/lif_mc.py
```diff
@@ -93,6 +93,7 @@ def forward(
                     dtype=input_tensor.dtype,
                 ),
             )
+            state.v.requires_grade = True
```
cpehle, line 96: Spelling error?

cpehle (changes requested):
LGTM (apart from typo). We should also adopt the convention used in LIFFeedForwardCell for the other Forward cells.

norse/torch/module/lif_mc.py
Jegp: Indeed. And unnecessary so I deleted it

cpehle (approved):

PR #128, opened 2020-10-04
Increase coverage

cpehle:


norse/torch/functional/test/test_threshold.py
```diff
@@ -1,8 +1,43 @@
...
 
 from norse.torch.functional.threshold import threshold, sign
 
 
+def test_threshold_throws():
+    alpha = 10.0
+    x = torch.ones(10)
+
+    with raises(ValueError):
+        out = threshold(x, "noasd", alpha)
```
Jegp, line 13: Codacy complains about `out` not being used. Replace with `_`?

norse/torch/functional/threshold.py
```diff
@@ -24,16 +24,16 @@ class HeaviErfc(torch.autograd.Function):
 
     @staticmethod
     def forward(ctx, x, k):
-        ctx.save_for_backward(x, k)
+        ctx.save_for_backward(x)
+        ctx.k = k
```
Jegp, line 28: Why store `k` if you're not using it?

Jegp (approved):
Looks good apart from the small linting issue. And I'm curious about the `k` in your tests.

norse/torch/functional/threshold.py
Jegp: Scratch that. I only saw an incomplete view

PR #129, opened 2020-10-07
Example fixes

Huizerd:
- `torch.cuda.is_available` instead of `torch.cuda_is_available`
- cartpole: write ANN and SNN to device to prevent CUDA error


cpehle (approved):
Thanks for these fixes! I think the old way of doing it was deprecated a while ago.

PR #131, opened 2020-10-07
Detached previous state variables to fix #130

### LINKED ISSUE #130: Reused state crashes subsequent backward passes
We're currently asking the state voltage to employ gradients for later backward passes. That works if the state is renewed for every cycle, but this crashes: 

```python

    model = LIFCell(12, 1)
    data = torch.ones(100, 12)
    out, s = model(data)
    loss = out.sum()
    loss.backward()
    out, _ = model(data, s)
    loss = out.sum()
```

Jegp:
Fixes #130 by checking whether state is assigned, and if so, ensures that the relevant state variables are detached


cpehle (changes requested):
This is not a good solution, it would break autograd and BPTT. You can't reuse the same computation graph twice (that is true), but this has nothing to do with the *Cell implementations. For use cases where you need to reuse the same graph twice you can actually specify it in the backward method (retain_graph = True). Otherwise you can also use (optimizer.zero_grad()) to reset the backward variables before calling backward again.

Jegp:
That's an excellent point! 
I think the problem is that the test cases are not sufficient. Let me elaborate.

Jegp:
Closed since this is not something Norse should worry about. If the computational graph breaks, the user needs to consider what happens to the gradients in the state themselves.

PR #132, opened 2020-10-07
Add further threshold tests, fix bugs

cpehle:
Also removed two unused files.


norse/torch/functional/stdp_sensor.py
```diff
@@ -38,7 +38,7 @@ def stdp_sensor_step(
     state: STDPSensorState,
     p: STDPSensorParameters = STDPSensorParameters(),
     dt: float = 0.001,
-) -> Tuple[torch.Tensor, STDPSensorState]:
+) -> STDPSensorState:
```
Jegp, line 41: Nice catch!

Jegp (approved):
Great! Nothing to say but LGTM

Issue #94, opened 2020-07-22
More Flexible Synapse Models


cpehle:
Currently we implement a limited number of synapse models. While this is sufficient for machine-learning oriented approaches it is limiting when it comes to neuroscience inspired approaches. Brian2 solves this by allowing users to specify synapse dynamics as described here https://brian2.readthedocs.io/en/stable/resources/tutorials/2-intro-to-brian-synapses.html. At a minimum we should look into defining some of the other common synapse types and provide a tutorial how to implement additional ones.

Jegp:
Sounds excellent! Is this a generalisation of #87?

cpehle:
To a degree it is. Currently we model synapses mostly through a weight (a floating point number) together with a fixed dynamic. Ideally we want to move away from that in two directions: 

- Each synapse should in principle carry more state than just it's weight.
- It should be possible for that state to have some temporal dynamic.

A synapse with STDP has both these additional properties and in addition also depends on more than just synaptic input from pre-synaptic neurons.

smalltimer:
Great suggestion. Having separate structures for synapses would certainly make things more intuitive for neuroscientists. E.g. in brian they are first-class citizens of the library just like neuronal layers. My suggestions to do this would be to:
1. Uncouple the synapses from layer-types (even the non-spiking torch layers)
2. Separating the FF and conv synapses as they imply very different calculation strategies

Jegp:
@smalltimer Thanks for the input. I think I'd agree with the separation, but I would also enjoy some form of hierarchy to ensure that it's still accessible for new users/easy to use. How about interpreting the synapse dynamics as a separate torch Module, which, in effect, would just be injected between the LIF layers to account for the summed weighting? If one wanted to have an entirely different synapse dynamics, it's simply a matter of putting a different module in sequence. 

In terms of examples, it could look like this:
```
Linear ->  LIF -> Linear -> LIF             # Regular weighted sum
STDP(Linear) -> LIF -> STDP(Linear) -> LIF  # STDP with linear weighting
SynapseX -> LIF -> SynapseY -> LIF          # Unknown, future synapse module
```

If that is to work, though, we'd need to be extremely clear in the nomenclature. It might make sense to still use the current LIFCell module, which is confusingly both a recurrent and linearly weighted LIF neuron.

ChauhanT:
[...]

PR #133, opened 2020-10-09
STDP steppers for FF and conv2d

ChauhanT:
Simulating one time-step for STDP FF LIF layers. This module updates both the post-synaptic layer, and the weights. There are 4 formulations of the STDP rule with varying levels of hardware-friendliness, and the option to impose hard bounds. 

A first draft, which could do with some python-specific optimisations.


norse/torch/functional/stdp.py
```diff
@@ -1,400 +1,276 @@
...
+    # Bound checking
+    w = p_stdp.bounding_func(w)
 
+    return (z_post, state_post, w, state_stdp)
 
-def conv2d_hard_multiplicative_stdp_step(
-    w: torch.Tensor,
+
+# %% Conv2D stepper
+def lif_conv2d_stdp_step(
```
Jegp, line 180: This is really neatly done. I can't help but wonder, though, if it can be generalised. If the approach depends on the neuron (`N`) and convolution (`C`), wouldn't that require `N * C` implementations?

Isn't there a more general approach where the same steps are taken, but in a general sense? One naïve suggestion (I'm not an STDP expert :-) ) could be to generalise the stdp part into a forward step and an update step. Something like this:
```python
def stdp_step(x: torch.Tensor, forward: Callable[...], update: Callable[...], p: STDPParameters, cell_state: SomeCellState):
    z, cell_state_post = forward(x, cell_state)
    new_state, p = update(x, z, p, cell_state, cell_state_post)
    return (z, new_state, p)
```
Could this make sense?

I imagine this is related to #94, since improved synapse abstractions would make it easier to operate on this.

Jegp (changes requested):
Really great work. The code seems thorough and clear. Two requests I would like to consider are
1. Improving the tests slightly. This is mostly historical, because we've been "bitten" before by overseeing dumb mistakes because we didn't check the numerical/structural values.
2. Looking into more general abstractions. I think the STDP idea is brilliant. I'm slightly concerned what this would mean if we are to scale this to all neuron models, for instance.

norse/torch/functional/stdp.py
Huizerd: I would say that the different neuron models are less of a problem than the `Linear` <-> `Conv2d`, especially given that things like #94 could make neuron models more alike. The problem separating `Linear` and `Conv2d` however is the receptive field of each neuron: while this is basically all previous neurons for the former, it is the convolutional filter for the latter, which requires some special ops. I guess it's OK if we have 2 implementations (linear and convolutional STDP) that work for all neurons

norse/torch/functional/stdp.py
```diff
@@ -1,400 +1,276 @@
-import torch
+# Includes
+from typing import Tuple
 
-from typing import Tuple, NamedTuple, Union
+import torch
+from torch import clamp, relu, pow as exponentiate
```
Huizerd, line 5: This import is confusing to me. I wouldn't import these separately, but just use `torch.pow`, `torch.clamp` and `torch.nn.functional.relu` (this is the same as `torch.relu`, but the code actually lives in `torch.nn.functional`, so this is more informative)

norse/torch/functional/stdp.py
```diff
@@ -1,400 +1,276 @@
...
+        z_pre,
+        z_post,
+        tau_pre_inv,
+        tau_post_inv,
+        a_pre,
+        a_post,
+        dt,
+    ):
+        self.t_pre += dt * tau_pre_inv * (-self.t_pre + a_pre * z_pre)
+        self.t_post += dt * tau_post_inv * (-self.t_post + a_post * z_post)
```
Huizerd, line 34: Don't in-place ops destroy gradient flow?

norse/torch/functional/stdp.py
```diff
@@ -1,400 +1,276 @@
...
-    ).sum(0).view(w.shape)
-    dw_minus = _A_minus_soft(w, p.w_min, p.eta_minus) * torch.bmm(
-        y_decayed.view(batch_size, out_channels, -1), z_pre_uf.permute((0, 2, 1))
-    ).sum(0).view(w.shape)
+    # STDP weight update
+    dw_plus = p_stdp.A_plus(w) * torch.einsum("bi,bj->ij", z_post, state_stdp.t_pre)
+    dw_minus = p_stdp.A_minus(w) * torch.einsum("bi,bj->ij", state_stdp.t_post, z_pre)
 
-    dw = dw_plus - dw_minus
+    w += dw_plus - dw_minus
```
Huizerd, line 171: Again, doesn't this destroy gradient flow?

Huizerd:
Nice work! I like how you merged the different STDP variants (soft and hard, additive and multiplicative), and how you made use of `einsum` and `nn.Linear`/`nn.Conv2d` to take care of most of the work. Some global comments:
- I would say that one of the benefits of the implementation I made before was that it would work with any neuron, because it only relied on the weight vector of that neuron and the type of connection. In your version you are constrained to the feedforward LIF, which is fine for an initial implementation, but I'm not sure I would throw away the old version for this
- Did you check mathematically that you get the correct results when using `einsum`/`Conv2d`/`Linear` ops?
- I always thought that in-place operations on tensors destroyed their gradient flow, but I'm not sure. Maybe worth checking

Again, really nice work, especially given your unfamiliarity with Python ;)

requirements-dev.txt
```diff
@@ -1,9 +1,10 @@
...
+-i https://pypi.org/simple
+astroid==2.4.2
+isort==5.6.1
+lazy-object-proxy==1.4.3
+mccabe==0.6.1
+pylint==2.6.0
+six==1.15.0
+toml==0.10.1
+typed-ast==1.4.1 ; implementation_name == 'cpython' and python_version < '3.8'
+wrapt==1.12.1
```
Huizerd, line 10: I would remove these changes

requirements.txt
```diff
@@ -1,5 +1,14 @@
-absl-py
-matplotlib
-numpy
-torch>=1.5.0
-torchvision>=0.6.0
+-i https://pypi.org/simple
```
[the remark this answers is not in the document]
Huizerd: I would remove these changes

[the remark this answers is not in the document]
ChauhanT: I used pipenv, and not very well, it seems. @Huizerd is absolutely right. The way the functions are written we shouldn't need any other headers besides norse. Will sort this out. Do you guys recommend using pipenv or just vanilla pip? pipenv is extremely convenient.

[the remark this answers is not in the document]
Huizerd: I always use `python3 -m venv venv`, which I think is included by default as a module in python distributions, or installed through `sudo apt install python3-venv`. See [here](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment). But whatever suits you, there are many virtual environment options. In any case, when you install norse and its requirements, these install many additional dependencies automatically. So your `requirements.txt` looks normal, but you don't have to commit it to the repo :)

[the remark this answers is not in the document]
ChauhanT: Aha. Thanks for the tip. As you can tell, it's the first time I'm doing a pull request to a non-personal repo. Of course, I should have only committed files that I'd like to merge into the repo...

norse/torch/functional/stdp.py
ChauhanT: Agreed. I thought quite a bit about this. Essentially linear and conv2d are two fundamentally different ways of calculating. 

Btw, I think we should also have an explicit conv2d spiking layer. Eventually the weight updates can simply be merged within the update step of the layers. It will take some doing though.

ChauhanT: Aha, @Jegp does make an excellent point though. Separating the two (stepping the layer and stepping the stdp) would make the implementation more general. I guess that was the idea behind @Huizerd 's initial implementation too.

norse/torch/functional/stdp.py
```diff
@@ -1,400 +1,276 @@
...
+        self.a_post = a_post
+        self.tau_pre_inv = tau_pre_inv
+        self.tau_post_inv = tau_post_inv
+        self.w_min = w_min
+        self.w_max = w_max
+        self.eta_plus = eta_plus
+        self.eta_minus = eta_minus
+        self.stdp_algorithm = stdp_algorithm
+
+        if self.stdp_algorithm == "additive":
```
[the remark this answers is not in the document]
smalltimer: As far as I understood, by using callable we will be forcing the user to link the right initialisation function to the stepper. I think this will make the STDP functionality less plug and play.

norse/torch/functional/stdp.py
smalltimer: Option#1  Removing the code for the lif stepper. This would mean the user needs to make sure they run the state updates before passing them on to the STDP stepper. It makes the library less user-friendly imo.

Option#2 A clean way to implement an integrated stepper would be if there were an FFState superclass (it can be an abstract class) which all layer-types inherit from. In C++ this is where one would specify the minimum required structure for the layer class, allowing the user to implement their own, library-compatible neurone models. I am sure there's an equivalent or better idiom for this in python!

I will implement option#1 for now.

ChauhanT: Option#1  Removing the code for the lif stepper. This would mean the user needs to make sure they run the state updates before passing them on to the STDP stepper. It makes the library less user-friendly imo.

Option#2 A clean way to implement an integrated stepper would be if there were an FFState superclass (it can be an abstract class) which all layer-types inherit from. In C++ this is where one would specify the minimum required structure for the layer class, allowing the user to implement their own, library-compatible neurone models. I am sure there's an equivalent or better idiom for this in python!

I will implement option#1 for now.

norse/torch/functional/stdp.py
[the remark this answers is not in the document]
ChauhanT: As far as I understood, by using callable we will be forcing the user to link the right initialisation function to the stepper. I think this will make the STDP functionality less plug and play.

norse/torch/functional/stdp.py
ChauhanT: Just checked. They do. Thanks for the heads up.

norse/torch/functional/test/test_stdp.py
```diff
@@ -1,113 +1,235 @@
...
-    s = STDPState(x=torch.zeros(1, pre), y=torch.zeros(1, post))
-
+        _, state_post, w, state_stdp = lif_linear_stdp_step(
+            z_pre[i], w, state_post, p_post, state_stdp, p_stdp
+        )
+    print("Additive:")
+    print(w)
+    print()
+
+    # Additive step
```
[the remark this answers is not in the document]
ChauhanT: Thanks for the tip. fixtures are great! Have added much more rigorous tests which check the shape as well as the calculations.

norse/torch/functional/stdp.py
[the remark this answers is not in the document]
Jegp: Just to make sure I understand @ChauhanT, would you say that it's too limiting to force the user to choose initialisation function in the STDPParameters case?

The current code looks conceptually pretty nice to me. I could still imagine factoring out the `stdp_algorithm` parameter to a function, so one would write, for instance `STDPParameters(..., Additive, ...)` instead of `STDPParameters(..., "additive", ...)`. But I'm also happy to leave it as-is :-)

norse/torch/functional/test/test_stdp.py
```diff
@@ -1,113 +1,232 @@
...
-    z_pre = torch.ones((time, batch, in_channels, *in_hw))
-    z_post = (torch.randn(time, batch, out_channels, *out_hw) > 0).float()
-    s = STDPState(
-        x=torch.zeros(batch, in_channels, *in_hw),
-        y=torch.zeros(batch, out_channels, *out_hw),
+def create_id(param):
+    return param[0]
+
+# Linear STDP
+@pytest.fixture(
```
Jegp, line 16: Wow, super advanced. Nice

Jegp (approved):
This looks pretty good to me. Robust, modular, and tested. Nicely done @ChauhanT!

I do have one comment about the implementation of `stdp_algorithm` as a string. But it's minor and I'd probably suggest merging this now and then we can iterate on that later :-)

norse/torch/functional/stdp.py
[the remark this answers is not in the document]
ChauhanT: @Jegp I see what you mean now, and I think from the code-base point of view you are absolutely right. Splitting it into small initialisation functions is probably better. I hadn't understood what 'callable' means in python. I'll look into this next week ;)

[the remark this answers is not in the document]
ChauhanT: @Jegp  I have been giving this some thought. The initialisations in the if...else use various combinations of the member attributes for different STDP algorithms. To factor these out will mean creating module-scope callables which have a uniform interface. I think it will make the code a little more modular, but less readable. 

If the idea is to encourage users to implement their own STDP algorithms, they will want to write much more than only the initialisation functions for the parameters (such users are likely to adopt a different parametrisation of the STDP dynamics altogether). 

So I would vote for leaving the if else cascade in there for now.

Jegp (changes requested):
Alright, just to bring the discussion to the main thread, I think we're nearing something we can merge and further iterate upon. @ChauhanT could you address @Huizerd's comments on making it general to not only work with LIF cells (looks like mostly a matter of naming, right?). 
Afterwards we can get some practical hands-on experience and then figure out how the code can improve. Does that sound ok? Also with you @Huizerd?

Huizerd:
Definitely!

ChauhanT:
Renamed the functions and pushed. I'm not sure how push requests work =.=" @Jegp  Please let me know what needs to be done next to get this merged!

Jegp (approved):
Looks fine. I'll wait for the checks to pass and then merge.
My only remaining comment would be that we can probably iterate on the comments, documentation, and usability of this going forward. 
Thanks for sticking with us so far @ChauhanT! :-)

Issue #116, opened 2020-09-24
Example of Pytorch Lightning training


cpehle:
Context: Pytorch lightning is an excellent tool, which makes it really easy to parallelise Pytorch models on clusters.

We should provide some example of how Norse is straightforwardly compatible with Pytorch lightning, since this sets it apart from other Pytorch based spiking neural network simulators.

Jegp:
Agreed. We could couple this with #88 where we can provide some really nice examples (and justifications) for specific models.

PR #134, opened 2020-10-10
Mobilenet-v2 port

cpehle:
This is a straight port of the Mobilenet-v2 vision model. It continues the work done in (#123). Next step is to actually test and tune training. Current plan is to introduce a torch lightning module to do that (#116).


Jegp (approved):
Nice work. I assume you tested that it learns already?

Also, why not go for pytorch-lightning?

cpehle:
> Nice work. I assume you tested that it learns already?
> Also, why not go for pytorch-lightning?
That would be next. I want to setup a training / evaluation pipeline for these kinds of large vision models.

PR #135, opened 2020-10-10
Add integration test for lif correlation

cpehle:
Tests the various correlation sensor related functions and modules.


Jegp (approved):
Looks nice. Closing in on that 100% mark!

PR #137, opened 2020-10-11
Fixed LSNN example in readme

Jegp:


cpehle (approved):
LGTM 👍, although we should really tackle #121 asap.

Issue #88, opened 2020-07-13
Improve model documentation


Jegp:
To help users engage with the library, understand use cases, and apply the code we should

* List available modules 
* Describe model characteristics / use cases
* Provide examples of uses
  * This could be links to existing code

Jegp:
Closing this quite abstract issue for now. We'll have to keep improving docs with specific tasks, like #160

PR #138, opened 2020-10-12
Added torch-like module/functional imports

Jegp:
In an effort to standardise our API (#121) and better document models (#88), this PR helps easen the code interface by allowing either
```python
import norse
norse.torch.LSNNCell
``` 
or
```python
from norse.torch import LSNNCell
```


cpehle (approved):
Good idea! 👍

Issue #76, opened 2020-06-17
Permit positional *and* key-value args in parametric module constructors


Jegp:
In the [LSNNLayer constructor](https://github.com/norse/norse/blob/2914199e2769da6f5bd5962e5536068f36815bc9/norse/torch/module/lsnn.py#L112) we permit positional arguments, but not key-value arguments. To improve useability, we should allow both as in: https://pytorch.org/docs/stable/_modules/torch/nn/modules/rnn.html#RNN

PR #139, opened 2020-10-14
Feature 121 consistent api

Jegp:
Solves #121 and #76 by streamlining *Cell and *Layer APIs and aligning with Torch conventions for imports.

Still need to 

- [x] Verify that all tasks and examples work
- [x] Verify that notebooks work


cpehle (approved):
Finally 👍, great job. Please make sure that the notebooks and example tasks also are still functional.

Issue #145, opened 2020-11-04
Fix BindsNet benchmark code


Jegp:
BindsNet discovered suboptimalities around their benchmark code. The benchmark code should be updated to be fair.

https://github.com/BindsNET/bindsnet/issues/425

PR #147, opened 2020-11-04
Fix 145 bindsnet

Jegp:
Addresses #145


norse/benchmark/norse_lif.py
```diff
@@ -22,46 +22,47 @@ def forward(
...
         return spikes
 
 
 def lif_feed_forward_benchmark(parameters: BenchmarkParameters):
-    model = LIFBenchmark(parameters).to(parameters.device)
-    input_spikes = PoissonEncoder(parameters.sequence_length, dt=parameters.dt)(
-        0.3
-        * torch.ones(
-            parameters.batch_size, parameters.features, device=parameters.device
+    with torch.no_grad():
```
cpehle, line 39: Have you checked whether this actually improves runtime performance?

norse/benchmark/main.py
```diff
@@ -55,6 +57,13 @@ def benchmark(
             for _ in range(config.runs):
                 duration = model(parameters)
                 durations.append(duration)
+                # Clean up by GC and empty cache
+                # Thanks to https://github.com/BindsNET/bindsnet/issues/425#issuecomment-721780231
+                gc.collect()
```
cpehle, line 62: I'm not sure if this is necessarily necessary or representative.

cpehle (changes requested):
Looks overall fine to me. Before finishing this, we should probably also update the benchmark README to point out that now BindsNET does not run out of memory.

norse/benchmark/main.py
Jegp: Unfortunately that's very hard to provide a clear answer to. It's normal in benchmarks that you 'prepare' the memory to provide as clean a slate as possible. Having a forced garbage collection does that. Whether it helps. Good question. But all in all, it should make it a tad more generalisable.

norse/benchmark/norse_lif.py
Jegp: Yeah, in my latest benchmarks it actually increases the performance by ~20% to leave the gradients out. Oddly enough, because I do remember we had conflicting earlier results.

cpehle (approved):
LGTM

PR #149, opened 2020-11-05
Optimization: torch.gt outperforms torch.where

### LINKED ISSUE #148: Threshold optimization
Now that we're talking about optimization, I saw that we use `torch.where` for the Heaviside step function that's used (among other things) for thresholds.

I did some quick benchmarks, and it seems that we could save some computation effort here by using `torch.ge`:

```python
import time
import random
import statistics

import torch


def where(x):
    return torch.where(x <= torch.zeros_like(x), torch.zeros_like(x), torch.ones_like(x))


def le1(x):
    return torch.le(x, 0.0).float()


def le2(x):
    return torch.le(x, torch.as_tensor(0.0)).float()


def le3(x):
    return torch.le(x, torch.zeros_like(x)).float()


functions = where, le1, le2, le3
times = {f.__name__: [] for f in functions}
runs = 100000
inputs = torch.randn(runs, 100, 100)

for i in range(runs):
    func = random.choice(functions)
    t0 = time.time()
    func(inputs[i])
    t1 = time.time()
    times[func.__name__].append((t1 - t0) * 1000)

for name, numbers in times.items():
    print('FUNCTION:', name, 'Used', len(numbers), 'times')
    print('\tMEDIAN', statistics.median(numbers))
    print('\tMEAN  ', statistics.mean(numbers))
    print('\tSTDEV ', statistics.stdev(numbers))
```

which, in my case, gives:

```
FUNCTION: where Used 25011 times
        MEDIAN 0.06723403930664062
        MEAN   0.06729976623091016
        STDEV  0.0014450391325887806
FUNCTION: le1 Used 25147 times
        MEDIAN 0.022411346435546875
        MEAN   0.022774781413584026
        STDEV  0.011103231984342292
FUNCTION: le2 Used 25005 times
        MEDIAN 0.021696090698242188
        MEAN   0.02181028943137154
        STDEV  0.000988286252607878
FUNCTION: le3 Used 24837 times
        MEDIAN 0.022172927856445312
        MEAN   0.022257421166154697
        STDEV  0.0007297306846966642
```

Of course, this is very much preliminary and I didn't consider any other factors such as gradient flow, JIT, etc, but it seems promising. Have you looked into this/were you aware of this?

Huizerd:
Closes #148

### DIFF
--- a/norse/torch/functional/heaviside.py
+++ b/norse/torch/functional/heaviside.py
@@ -10,8 +10,4 @@ def heaviside(data):
     .. math::
         H[n]=\\begin{cases} 0, & n <= 0, \\ 1, & n \\g 0, \\end{cases}
     """
-    return torch.where(  # pragma: no cover
-        data <= torch.zeros_like(data),
-        torch.zeros_like(data),
-        torch.ones_like(data),
-    )
+    return torch.gt(data, torch.as_tensor(0.0)).to(data.dtype)


norse/torch/functional/heaviside.py
```diff
@@ -10,8 +10,4 @@ def heaviside(data):
     .. math::
         H[n]=\\begin{cases} 0, & n <= 0, \\ 1, & n \\g 0, \\end{cases}
     """
-    return torch.where(  # pragma: no cover
-        data <= torch.zeros_like(data),
-        torch.zeros_like(data),
-        torch.ones_like(data),
-    )
+    return torch.gt(data, torch.as_tensor(0.0)).to(data.dtype)
```
Jegp, line 13: And the code is even shorter! Win-win :)

Jegp (approved):
Short and sweet. Nice work!

cpehle (approved):

PR #150, opened 2020-11-06
First attempt at more flexible synapse models #94

Jegp:
Let's start a discussion on how to approach #94. This first attempt is trying to build a first, generic synapse model that can lay the ground for more general models. I added a crude STDP example to illustrate how this could work.


Jegp:
Closed due to deprecation

PR #152, opened 2020-11-15
Added functional lift and corrected lif jit

Jegp:
Added a functional lift that takes a higher-order activation function and runs it in time.

Useful for injecting activation functions without using the module API.

I also fixed a minor problem in the lif_jit where batched inputs wouldn't provide the correct output shape in the first take (with the default state), because broadcasting aren't taking place in the first pass of the lif.


cpehle (approved):
Nice idea 👍.

Jegp:
I'm slightly hoping we can use this as a basis for a C++ optimisation where we can offload the loop in a nice and generic way. TBD.

PR #158, opened 2020-11-21
Model documentation

Jegp:
* Retrofitted the documentation to same structure as main page\
* Added descriptions on how to get started with Norse
* Minor fixes and typos

### DIFF
--- a/docs/source/hardware.rst
+++ b/docs/source/hardware.rst
@@ -1,7 +1,7 @@
 .. _page-hardware:
 
 Hardware acceleration
--------------------
+---------------------
 
 Norse is built on top of `PyTorch <https://pytorch.org>`_ which has excellent support for hardware acceleration. 
 This article will cover how Norse can be accelerated with GPUs through the `CUDA platform <https://en.wikipedia.org/wiki/CUDA>`_.
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,8 +1,10 @@
+.. page-index:
+
 .. norse documentation master file, created by
    sphinx-quickstart on Mon Jun 24 16:20:44 2019.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
-   
+
 
 Norse
 =====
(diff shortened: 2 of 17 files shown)


docs/source/hardware.rst
```diff
@@ -1,7 +1,7 @@
 .. _page-hardware:
 
 Hardware acceleration
--------------------
+---------------------
```
cpehle, line 4: nice this has been bothering me for a while....

docs/source/index.rst
```diff
@@ -1,8 +1,10 @@
+.. page-index:
```
cpehle, line 1: Do you know what these GitHub Actions mean?

docs/source/learning.rst
```diff
@@ -3,9 +3,81 @@
...
+
+Training Spiking Neural Networks
+=====================================
+
+Many solutions have been attempted to solve this problem with varying 
+success. 
+Here we will only cover the *surrogate gradient* approach, and
+illustrate how  it is implement in norse.
+
+Our approach builds on the SuperSpike method proposed by
```
cpehle, line 31: This is not entirely true. To the best of my knowledge it was first introduced by https://www.pnas.org/content/113/41/11441

cpehle: There is a review by Neftci, Zenke and Mostafa of "Surrogate Gradient methods"

cpehle (approved):
Really nice work overall, we should be a bit more careful with attributing the methods, maybe we should start a collection of references.

docs/source/index.rst
Jegp: I would think GitHub is clever enough to link warnings/errors to the code. This is probably one... Of many. I only cleaned up ~30% of the errors. There are still so many left. Especially from absail.

cpehle: Yeah abseil, wasn't a great idea....

Jegp: Well. It's a start. Did you see the argparse flags that Pytorch Lightning introduces? I really like that approach

Issue #163, opened 2020-11-30
There is a problem when I use LSNN


guoweiyu:
The model i used is like this:
model = SequentialState(
    norse.LSNNLayer(norse.LSNNCell,12, 32),
    norse.LSNNLayer(norse.LSNNCell,32, 64),
    norse.LSNNLayer(norse.LSNNCell,64, 64),
    norse.LSNNLayer(norse.LSNNCell,64, 32),
    norse.LSNNLayer(norse.LSNNCell,32, 10),
    nn.Flatten(),
    norse.Lift(torch.nn.Linear(2000, 256)),
    norse.Lift(torch.nn.Linear(256, 10))
)
        read_out,state=model(x.cuda(),state)
        optimizer.zero_grad()  
        loss.backward()

the error:one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [10, 10]], which is output 0 of TBackward, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).

when i change the code like: read_out,state=model(x.cuda()). The code works

But I want state to be passed between multiple inputs. How could i fix this question.

Jegp:
Hi @guoweiyu, thanks for writing us! Did you try to use the latest master branch? I think this was fixed with #127 
```bash
pip install git+https://github.com/norse/norse
```

guoweiyu:
Thanks a lot for your reply!
But it is still not working.
I write a simple test for this questions.  I get the same error when i run it. Maybe you can try to run it. Thanks again.


import torch
import norse.torch as norse
from norse.torch.module.sequential import SequentialState
from torch import nn
model = SequentialState(
    norse.LSNNLayer(norse.LSNNCell,12, 32),
    norse.LSNNLayer(norse.LSNNCell,32, 64),
    norse.LSNNLayer(norse.LSNNCell,64, 64),
    norse.LSNNLayer(norse.LSNNCell,64, 32),
    norse.LSNNLayer(norse.LSNNCell,32, 10),
    nn.Flatten(),
    norse.Lift(torch.nn.Linear(2000, 256)),
    norse.Lift(torch.nn.Linear(256, 10))
)
optimizer = torch.optim.Adam(model.parameters(), lr=0.0001)
loss_func = nn.MSELoss()
data = torch.ones(3,16, 200, 12)
target=torch.ones(16,10)
state=None
for x in data:
[...]

PR #164, opened 2020-11-30
Broken backprop with sequential state

cpehle:
Using SequentialState with LSNNLayer doesn't work (#163). This pull request aims to fix this.


Jegp:
Nice work. May I suggest you reduce the number of layers and neurons? Theoretically, we should be able to get the same results with less computation.

Jegp (approved):
LGTM

PR #176, opened 2021-01-12
Removed autopses by default in neuron modules. Closes #19

### LINKED ISSUE #19: Remove autapses from recurrent modules
Currently the recurrently connected spiking neural network modules allow autapses (self connections). This might not be a sensible default.

Jegp:
I set autopses to be **disabled** by default.


cpehle (approved):
LGTM

PR #177, opened 2021-01-13
Izhikevich model

cpehle:
Implement the functional part of the Izhikevich model according to the original Matlab code provided at https://www.izhikevich.org/publications/spikes.htm. We should investigate a different solver.


norse/torch/functional/izhikevich.py
```diff
@@ -0,0 +1,159 @@
...
+    s = IzhikevichState(v=-63.8 * torch.ones(shape), u=-63.8 * p.b * torch.ones(shape))
+    return s, p
+
+
+def izhikevich_step(
+    input_current: torch.Tensor,
+    s: IzhikevichState,
+    p: IzhikevichParameters,
+    dt: float = 0.001,
+):
```
Jegp, line 151: A few lines of documentation would be really helpful. Perhaps just to describe what the model is and where to read more about it.

norse/torch/functional/test/test_izhikevich.py
```diff
@@ -0,0 +1,464 @@
...
+mpl.rcParams["xtick.bottom"] = False
+mpl.rcParams["xtick.labeltop"] = False
+mpl.rcParams["xtick.labelbottom"] = False
+mpl.rcParams["ytick.left"] = False
+mpl.rcParams["ytick.right"] = False
+mpl.rcParams["ytick.labelleft"] = False
+mpl.rcParams["ytick.labelright"] = False
+
+
+def plot_test():
```
Jegp, line 26: This test doesn't seem to assert anything. Would it make sense to put it somewhere else so it doesn't get picked up by pytorch? We could start a plotting module or something similar since we'd like to start that effort soon anyway.

norse/torch/functional/test/test_izhikevich.py
```diff
@@ -0,0 +1,464 @@
...
+        us.append(s.u)
+        if t > T1:
+            input_current = 14 * torch.ones(1)
+        else:
+            input_current = 0 * torch.ones(1)
+        _, s = izhikevich_step(input_current, s, p)
+        cs.append(input_current)
+
+    plt.subplot(5, 4, 1)
+    plt.plot(vs)
```
Jegp, line 45: Does this actually work in a pytest context? I've never tried that before :)

norse/torch/functional/test/test_izhikevich.py
Jegp: It's also a reeeeeaaly long function ;)

Jegp (changes requested):
Super nice and comprehensive work!
It'd be nice to have documentation. And perhaps a module equivalent.

norse/torch/functional/test/test_izhikevich.py
cpehle: Yes this was actually not really meant as a test but just as a way to figure out, whether the neuron did roughly what I expected it to do.

Jegp: And the code is relevant. I'm just wondering if we can exclude it from pytest's search pattern. This could actually likely be achieved by dropping the `test_` prefix

Issue #166, opened 2020-11-30
Explore single RNN-like module for *all* neuron dynamics


Jegp:
PyTorch has a single [RNN base module](https://pytorch.org/docs/stable/_modules/torch/nn/modules/rnn.html#RNNBase) that applies all the recurrent activation dynamics in time. We should first study how the module works (why are they using strings as activations and not callables?!) and how they shove gradients around (#163) and then build a similar construction for Norse as discussed in #121.

Here's an ad hoc checklist:
- [x] Build a Norse version of the RNN Base
- [x] Add thorough tests for functionality
- [x] Add thorough tests for gradients in *multiple* connected layers

Jegp:
This is also interesting for inspiration: https://keras.io/api/layers/recurrent_layers/time_distributed/

Huizerd:
I have an idea for this, which would also tackle #169 and #143 (and #29).

Not sure it's fully working, but most of it should be. I think it decreases the 'thickness' of the layer on top of PyTorch that is Norse, and gives the user a bit more flexibility while still being useful. I also implemented heterogeneous/learnable time constants. Please have a look!

```python
from typing import NamedTuple, Optional, Tuple

import torch
import torch.nn as nn


class SpikeFunction(torch.autograd.Function):
    @staticmethod
    def forward(ctx, v):
        ctx.save_for_backward(v)
        return v.gt(0.0).float()

    @staticmethod
    def backward(ctx, grad_output):
        (v,) = ctx.saved_tensors
        grad_input = grad_output.clone()
        spike_pseudo_grad = 0.3 * torch.nn.functional.relu(1 - v.abs())
        return grad_input * spike_pseudo_grad, None


def threshold(v: torch.Tensor, *args) -> torch.Tensor:
    return SpikeFunction.apply(v)


class OptionalLearnable(NamedTuple):
    tensor: torch.Tensor
    learnable: bool


class LIFParameters(nn.Module):
    def __init__(
        self,
        tau_syn_inv: OptionalLearnable = OptionalLearnable(tensor=torch.tensor(1.0 / 5e-3), learnable=False),
        tau_mem_inv: OptionalLearnable = OptionalLearnable(tensor=torch.tensor(1.0 / 1e-2), learnable=False),
        v_leak: torch.Tensor = torch.tensor(0.0),
        v_th: torch.Tensor = torch.tensor(1.0),
        v_reset: torch.Tensor = torch.tensor(0.0),
        method: str = "super",
[...]

PR #178, opened 2021-01-24
First iteration of SNN module restructuring

Jegp:
This PR attempts to attack a restructuring of the module interfaces (#166). It is flawed in a number of ways:

* Inconsistency between *Layer, *FF, *Cell modules
* A lot of code duplication
* Divergent code means difficulty to optimize

I have tried to
* Reduce the module interfaces to **three base classes**
  1. `FeedforwardSNN` which represents **shapeless** (no input/hidden shape) and **timeless** feedforward dynamics (also fixed the `FeedForward` misspelling)
  2. `RecurrentSNNCell` which represents **recurrent** and **timeless** dynamics (equivalent to PyTorch's `RNNCell`)
  3. `RecurrentSNN` which represents **recurrent** and **timed** dynamics (equivalent to PyTorch's `RNNBase`/`RNN`)

I would still like to
-  ~[ ] the state to a single tensor instead of a `NamedTuple`~
-  ~[ ] Add the parameters as modules (as suggested by @Huizerd)~
- [x] Migrate LSNN
- [x] Migrate LIF variants (Ex, AdEx, etc.)

What do you think so far? Contributions are more than welcome! :-)


norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+            recurrent_weights if autopses else remove_autopses(recurrent_weights)
+        )
+
+    def forward(self, input_tensor: torch.Tensor, state: torch.Tensor):
+        state = state if state is not None else self.state_fallback(input_tensor)
+
+        T = input_tensor.shape[0]
+        outputs = []
+
+        for ts in range(T):
```
Huizerd, line 190: I wondered about this the other day: you can either have a time loop per neuron/layer (as done here), or you can have only one for the entire network. In the former you would go through all time steps for a single layer and then move on to the next and go through all steps for that one etc.; in the latter you would only have one time loop, but for each iteration you would go through all layers.

I was wondering whether there would be a different outcome for these?

norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+        p (torch.nn.Module): The neuron parameters as a torch Module, which allows the module
+            to configure neuron parameters as optimizable.
+        dt (float): Time step to use in integration. Defaults to 0.001.
+    """
+
+    def __init__(
+        self,
+        activation: FeedforwardActivation,
+        state_fallback: Callable[[torch.Tensor], torch.Tensor],
+        p: torch.nn.Module,
```
Huizerd, line 42: I think this would allow the learnable parameters yes

norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+]
+
+RecurrentActivation = Callable[
+    # Input        State         Input weights Rec. weights  Parameters       dt
+    [torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.nn.Module, float],
+    Tuple[torch.Tensor, torch.Tensor],
+]
+
+
+class FeedforwardSNN(torch.nn.Module):
```
Huizerd, line 23: Not sure about the name, but that's probably for later. It's more of a `Cell` as well, I think

norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(
+        self,
+        input_size: int,
+        hidden_size: int,
+        activation: FeedforwardActivation,
+        state_fallback: Callable[[torch.Tensor], torch.Tensor],
+        p: torch.nn.Module,
+        input_weights: Optional[torch.Tensor] = None,
+        recurrent_weights: Optional[torch.Tensor] = None,
+        autopses: bool = False,
```
Huizerd, line 89: Shouldn't this be autapse?

norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+            self.input_weights = torch.nn.Parameter(
+                torch.randn(hidden_size, input_size) * torch.sqrt(2 / hidden_size)
+            )
+        if recurrent_weights is not None:
+            self.recurrent_weights = recurrent_weights
+        else:
+            self.recurrent_weights = torch.randn(hidden_size, hidden_size) * torch.sqrt(
+                2 / hidden_size
+            )
+        self.recurrent_weights = torch.nn.Parameter(
```
Huizerd, line 110: I think you can do:
```python
        if recurrent_weights is not None:
            self.recurrent_weights = recurrent_weights
        else:
            self.recurrent_weights = nn.Parameter(
                torch.randn(hidden_size, hidden_size) * torch.sqrt(2 / hidden_size)
            )

        if not autapses:
            with torch.no_grad():
                self.recurrent_weights.fill_diagonal_(0.0)
```

norse/torch/module/snn.py
```diff
@@ -0,0 +1,201 @@
...
+            remove autopses in custom recurrent weights, if set above.
+        dt (float): Time step to use in integration. Defaults to 0.001.
+    """
+
+    def __init__(
+        self,
+        input_size: int,
+        hidden_size: int,
+        activation: FeedforwardActivation,
+        state_fallback: Callable[[torch.Tensor], torch.Tensor],
```
Huizerd, line 85: I like this, would make future code changes regarding state easier

norse/torch/module/snn.py
Jegp: I think we should facilitate both. And theoretically, the outcome should be the same. The first case is exactly taken care of by the above, but to do the second case, you'd need something that groups all the dynamics into a single module and runs them over all the timesteps. I think this is pretty much what @cpehle tried to do in the [Lift module](https://github.com/norse/norse/blob/master/norse/torch/module/lift.py#L57).

The question we should be asking ourselves, I guess, is whether there is a better way of achieving this. Should we, for instance, augment the [SequentialState](https://github.com/norse/norse/blob/master/norse/torch/module/sequential.py) to also support time? Can we think of something better? Or should we not include these abstractions at all and let people code this themselves?

norse/torch/module/snn.py
Jegp: @cpehle also just had the idea that we could technically fuse the FF modules with the Cell modules if, for instance, allowed to set the weights to `None`. 

One option could also be to remove the non-temporal recurrent layer, so we could simply have an `SNNCell` and an `SNN` module. That would be much more in line with the PyTorch nomenclature. But I wonder if someone would be missing the non-temporal recurrent dynamics. For instance, in the case of running a sequence of modules for every timestep.

norse/torch/module/snn.py
Jegp: Don't know. We also write weight**s**. Theoretically, it is about killing *all* the autopse**s** between all the neurons in the module. But would it be easier to understand in singular?

Jegp: Ah, you meant the spelling :laughing:! Yes, that completely true! But would you also do singular?

norse/torch/module/snn.py
Jegp: Awesome comment! That pretty much removes ~30LOC. Thanks, @Huizerd!

Jegp:
Two items I'd like your feedback on:

* I'm considering dropping the linear part of the recurrent modules. Shouldn't it be up to the user to specify whether the input should be transformed?
* @cpehle had the idea that we could merge `SNNCell` and `FeedforwardSNNCell` into one layer, and parameterize the recurrence. It's pretty much the same for the user, and we could avoid confusing nomenclature (feedforward cell vs. just cell). Especially if we drop the linear part above.

norse/torch/module/snn.py
Huizerd: No plural is fine, I just meant the 'a' :smile:

Huizerd:
>
>     * I'm considering dropping the linear part of the recurrent modules. Shouldn't it be up to the user to specify whether the input should be transformed?
> 
>     * @cpehle had the idea that we could merge `SNNCell` and `FeedforwardSNNCell` into one layer, and parameterize the recurrence. It's pretty much the same for the user, and we could avoid confusing nomenclature (feedforward cell vs. just cell). Especially if we drop the linear part above.
>

Yes, I agree. Users would provide an input connection, e.g. `nn.Linear`, themselves, and optionally another `nn.Linear` (or something else) if they also want recurrency. So you would only need one base class, e.g. `SpikingCell` for pretty much any neuron type, feedforward or recurrent. There could then be another convenience class with the time loop like you had, e.g. `SpikingLayer`, such that we can have both the loop options you/I mentioned in one of the comments.

Huizerd:
Such that Norse only provides those things that people find most difficult about SNNs: dynamics and temporal loops

Jegp:
> Yes, I agree. Users would provide an input connection, e.g. `nn.Linear`, themselves, and optionally another `nn.Linear` (or something else) if they also want recurrency. So you would only need one base class, e.g. `SpikingCell` for pretty much any neuron type, feedforward or recurrent. There could then be another convenience class with the time loop like you had, e.g. `SpikingLayer`, such that we can have both the loop options you/I mentioned in one of the comments.

I actually quite like the idea of a `Linear` layer as recurrence. I initially thought to add a `recurrence: Union[bool, Tensor]` parameter because we want to describe three situations: no recurrence, recurrence with defaults, recurrence with specifics. But it's confusing and in your situation we could actually also add recurrent bias. It would then look something like this:
```python
LIFCell(2, 4, recurrence = None)            # 1. No recurrence (default?)
LIFCell(2, 4, recurrence = nn.Linear(4, 4)) # 2. Default recurrence
linear = nn.Linear(4, 4)                    # 3. Weighted recurrence
linear.weight = ...
LIFCell(2, 4, recurrence = linear)
```
The downside of this would be that the kaiming distribution of weights in the linear layer might not be optimal. But the upside is that we can just add any torch module to describe the recurrent dynamics.

I would also like to add a "temporal" layer, just to follow the Torch conventions. Such that we have `SNNCell`/`LIFCell` (w/o time) and `SNN`/ LIF` (with time). Does that sound 'digestible' to users?

Jegp:
Alright, I've given it another iteration. I liked the way you phrased the "responsibilities" of Norse @Huizerd: dynamics and temporal loops. At the same time I'm withdrawing on some of the more structural changes we talked about for two reasons: time and effort. Particularly, I think it requires a bit more careful thought to handle the recurrent parameters and find a proper way to implement the recurrence as modules. 

To avoid crippling complexity I'd suggest we settle on a user-facing class API and stick to the checklist I wrote in the initial PR. We can later adapt the interface as long as the main classes are in place.

Currently, I would suggest that we

1. Split non-recurrent neuron dynamics and time: `LIFCell` (without time) `LIF` (with time)
   * This aligns with PyTorch's API and is straight-forward to extend.
2. Split modules between recurrent and non-recurrent modules: `LIFRecurrent` vs. `LIF`
   * Getting the logic and parameters right is not something we should take too lightly I think
   * Furthermore, thinking about the recurrent dynamics, it might be better practice to split it up into two modules, since the recurrent neurons are logically separate. It also makes it easier to optimize stand-alone modules instead of dealing with logic inside the modules.

In sum, that would mean 4 modules for each neuron type: 
1. `LIFCell` (previous feedforward cell)
2. `LIFRecurrentCell` (did not exist before)
3. `LIF` (wraps the `LIFCell` in time)
4. `LIFRecurrent` (wraps the `LIFRecurrentCell` in time)

Huizerd:
Ok, after some more consideration it indeed seems more natural to have separate feedforward and recurrent classes. Now I'm thinking about the activation functions that are also different for feedforward and recurrent. Given that in the case of a recurrent neuron, the feedforward and recurrent inputs are just added together, could we have the following signature for an activation function:

```python
def activation(inputs: torch.Tensor, state: torch.Tensor, p: torch.nn.Module, dt: float) -> torch.Tensor:
    ...
    return outputs
```

and call it like this:

```python
# Feedforward
self.activation(inputs, state, self.p, self.dt)

# Recurrent option 1
inputs = self.input_weights @ inputs + self.recurrent_weights @ state.z
self.activation(inputs, state, self.p, self.dt)
# Recurrent option 2
self.rec_wrapper(self.activation, inputs, state, self.input_weights, self.recurrent_weights)

def rec_wrapper(activation: Callable, inputs: torch.Tensor, state: torch.Tensor, ff_weights: torch.Tensor, rec_weights: torch.Tensor) -> torch.Tensor:
    inputs = ff_weights @ inputs + rec_weights @ state.z
    return activation(inputs, state, self.p, self.dt)
```

Just to put this out there :)

Jegp:
I think that's a pretty good idea actually. And when you think about it, that doesn't change the user-facing API at all. It's just a matter of how we choose to structure the code internally!

If one were a complete OOP nerd, one could even imagine having a single class/function that implements the activation function, and then inherit into the other classes. I'm not certain it makes sense, but it could work. And it would remove the ludicrous code duplication errors from PyLint :facepalm:

Jegp:
Oh, and I deeply appreciate all the feedback @Huizerd btw. Thanks a lot!

Huizerd:
> If one were a complete OOP nerd, one could even imagine having a single class/function that implements the activation function, and then inherit into the other classes. I'm not certain it makes sense, but it could work. And it would remove the ludicrous code duplication errors from PyLint

Yes, that sounds like a good idea! And no problem, happy to help :)

docs/source/conf.py
```diff
@@ -42,16 +42,16 @@
     "sphinx.ext.napoleon",
     "sphinx.ext.githubpages",
     "sphinx_autodoc_typehints",
-    "sphinx_gallery.gen_gallery",
+    # "sphinx_gallery.gen_gallery",
```
cpehle, line 45: remove entirely? I still like the idea of examples in the documentation, but maybe there is a better way.

docs/source/conf.py
```diff
@@ -42,16 +42,16 @@
...
 bibtex_bibfiles = ['bibliography.bib']
 
-sphinx_gallery_conf = {
-    "examples_dirs": "../../norse/examples",  # path to your example scripts
-    "gallery_dirs": "auto_examples",  # path to where to save gallery generated output
-}
+# sphinx_gallery_conf = {
+#     "examples_dirs": "../../norse/examples",  # path to your example scripts
+#     "gallery_dirs": "auto_examples",  # path to where to save gallery generated output
+# }
```
cpehle, line 54: See above

README.md
```diff
@@ -116,19 +116,20 @@ This classifier is a taken from our [tutorial on training a spiking MNIST classi
...
 import torch, torch.nn as nn
 from norse.torch import LICell             # Leaky integrator
-from norse.torch import LIFFeedForwardCell # Leaky integrate-and-fire
+from norse.torch import LIFCell            # Leaky integrate-and-fire
 from norse.torch import SequentialState    # Stateful sequential layers
 
 model = SequentialState(
     nn.Conv2d(3, 20, 5, 1),      # Convolve from 3 -> 20 channels
-    LIFFeedForwardCell(),        # Spiking activation layer
+    LIFCell(),                   # Spiking activation layer
```
cpehle, line 124: Inconsistent with what you are writing in getting started below.

docs/source/started.rst
```diff
@@ -73,7 +73,7 @@ Luckily, we offer our own `SequentialState <https://norse.github.io/norse/auto_a
 
     model = SequentialState(
         torch.nn.Linear(10, 5),
-        norse.LIFFeedForwardCell(),
+        norse.LIF(),
```
cpehle, line 76: Is it LIFCell or LIF? See above.

norse/torch/functional/lif.py
```diff
@@ -22,13 +22,24 @@ class LIFParameters(NamedTuple):
...
+        alpha: float = torch.as_tensor(100.0),
+    ):
+        super().__init__()
+        self.tau_syn_inv = tau_syn_inv
+        self.tau_mem_inv = tau_mem_inv
+        self.v_leak = v_leak
+        self.v_th = v_th
+        self.v_reset = v_reset
+        self.method = method
+        self.alpha = alpha
```
cpehle, line 42: I thought in this changeset we wouldn't switch over to torch.nn.Module for the parameters.

cpehle (changes requested):
I think we can merge these changes as is pending the few minor changes. We should then iterate based on your discussion in this pull request + issue on the outstanding changes. Overall great work and I think this is a good improvement!

docs/source/conf.py
Jegp: I completely agree. At the same time I'm trying to really reduce the number of things we have to maintain. I'm also happy to leave it as-is. But as it stands, we only have one example which is a bit of a contrast to the tasks and notebooks, which pretty much have the same function.

norse/torch/functional/lif.py
Jegp: Indeed, this is a remnant of the previous example. Thanks for pointing this out!

README.md
Jegp: Good catch. I haven't fully looked through the docs. I'll go through it now

cpehle (approved):
LGTM for now, should probably iterate more to address the other points raised in the discussion. But this is a great start!

docs/source/started.rst
Jegp: I updated the documentation (mainly getting started and working with pages). It's actually easier to explain, which I guess is a good sign.

cpehle: Yeah I think that is a good sign!

Jegp:
Great. I also like where we're heading, but there could be a few things I missed.

I'll make sure the static code analysis and coverage looks ok and then we can merge. Thanks for your feedback @cpehle and @Huizerd!

docs/source/conf.py
cpehle: Sure we should't keep stale content, I think it was the first thing I wrote to explain the library and it is kind of nice to have tutorial style content in the docs, but we should maybe stick to well maintained notebooks for now. One advantage of these examples is that they automatically get build and executed as part of the build process so it is easy to detect failures and they can be exported / opened as Jupyter notebooks.

cpehle: In other words they should never be stale relative to the actual state of the library.

PR #185, opened 2021-01-31
Bumped version

Jegp:


Jegp:
Are we ready for releasing? Is there anything we should include in the issue list? Otherwise, I'll just push all the .6 issues to a .7 milestone.

cpehle (changes requested):
Can you rebase on top of the two changes I made, currently tests are failing on my Mac setup and I don't think they introduce regressions.

cpehle (approved):

PR #188, opened 2021-02-18
Fix LSNNRecurrent Readme example

cpehle:
Maybe there is a way to automatically check these examples? In any case we should
make sure to check them, when API changes are made.


Jegp:
Yeah, that's not good. I thought I tested it. Thanks for catching this!

Jegp (approved):

Issue #156, opened 2020-11-20
Add visualisation tools


Jegp:
It would be helpful to have visualisation tools to plot/debug information in the SNNs. I also think we should be slightly careful to use correct/adequate abstractions since it is 1) not the primary purpose of Norse, and 2) we don't want to maintain something that will change a lot in the future.

Here are a few suggestions for visualisations: 
- Layer parameters
  - Weights / tuning curves
  - Membrane voltages/currents
  - Both over time and at a single time-step
- Spike raster plots (from one or more layers)

Any other interesting and useful visualisations we'd need?

Looper2074:
It would be nice to have a visualization like this:![1_VYAr7eMJfGLCyd60Z-YjdA.png](https://user-images.githubusercontent.com/58263228/103542024-f9f33200-4ea4-11eb-9557-f01c682b49f7.png)

https://towardsdatascience.com/saliency-map-using-pytorch-68270fe45e80

Jegp:
Indeed @Looper2074. But we have to be slightly careful in that I wouldn't want to develop too specific visualization tools for Norse. Mostly because that's not the reason for Norse in the first place. Come to think of it, what could make even more sense is to develop an independent visualization tool for event-based data in general. That would be fascinating and highly relevant, I think.
The difficult part would then be to figure out what is fitting/reasonable to have *within* Norse and what should be moved to a separate tool/library. Somehow I can actually imagine the above suggestions to exist in a separate repository...

Jegp:
Hi @Looper2074,
I've finally merged #193 which adds plotting primitives as well as examples on how to plot. They are still *primitives* in the sense that they are building blocks you'll have to use to your ends. But I think they should help you to achieve at least a part of what your want. In your example, for instance, you can now write the following:

```python
from norse.torch import plot_heatmap_2d
saliency = ...
plot_heatmap_2d(saliency, cmap="hot")
```

That said, there is still some work to make it prettier and more modular. If you have any ideas about how this can be done well, I'm eager to hear from you.

For now, I'm closing the issue. Feel free to reopen if you think there's something missing.

PR #193, opened 2021-02-28
Added features to diagnose and plot

Jegp:
Addresses #156


Jegp:
If there are no further comments on this, I'll merge it in a day or two :-)

cpehle (approved):
Very nice work overall 👍. Hopefully we have some good examples of their use soon.

PR #194, opened 2021-03-04
Modifications to Izhikevich tests

adelpierre:
Modified the `test_izhikevich.py` to fix [this check failure](https://github.com/norse/norse/pull/177)


cpehle (approved):
Thanks for cleaning this up, I think it would be worthwhile to add functional tests as a next step. The Izhikevich model is also a good example for a model which could benefit from a more sophisticated integration scheme. So we can use it as a testing ground for the corresponding issues.

PR #195, opened 2021-03-15
Additions to the Izhikevich model

adelpierre:
Adresses #177 : creation of an Izhikevich module containing a working IzhikevichCell (+ WIP on IzhikevichLayer), addition of documentation and changes to functions for easier calling in the module


norse/torch/functional/izhikevich.py
```diff
@@ -0,0 +1,173 @@
...
+    Parameters:
+        p (IzhikevichParameters) : parameters of the Izhikevich neuron model
+        s (IzhikevichState) : state of the Izhikevich neuron model
+    """
+
+    p: IzhikevichParameters
+    s: IzhikevichState
+
+
+p_ts = IzhikevichParameters(a=0.02, b=0.2, c=-65, d=6)
```
Jegp, line 59: I can see why you'd want to have the parameters stored like this, but this means that the `p_ts` variable gets exposed in the `izhikevich.py` module upon import. If you'd like to keep it this way, I'd suggest you rename it to something more meaningful. Like `tonic_spiking_p`. That's a bit easier to read than `ts`.

Jegp: Same comment applies to the cases below.

norse/torch/module/izhikevich.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class IzhikevichCell(SNNCell):
+    """Documentation WIP
+    Module that computes a single Izhikevich neuron-model *without* recurrence and *without* time.
+
+    Parameters:
+        spiking_method (IzhikevichSpikingBehaviour) : parameters and initial state of the neuron
+    Examples:
+        >>> batch_size = 16
```
Jegp, line 27: Excellent you added examples! One good practice to follow is to ensure your examples are [SSCCE](http://www.sscce.org/). The only thing you'd need is to add import statements. C.f. my comment about the `__init__` file.

Jegp (changes requested):
Nice work! I have a few inline comments and then I would also like you to have a look at the `norse/torch/functional/__init__.py` and `norse/torch/module/__init__.py`. Both files exposes components from the `functional` and `module` packages. You'd like want something similar for the Izhikevich packages. Basically, you'd "just" need to import then and then add them as a `str` to the `__all__` list. Yeah, it's a weird property of Python. What this allows you to do is basically to write `from torch.norse import tonic_spiking` or `import torch.norse as snn; snn.tonic_spiking`. Which is much easier than `import norse.torch.functional.izhikevich.tonic_spking`.

I'm of course happy to hear suggestions if you have better ideas :-)

adelpierre:
I think I've addressed the comments you made, adding the components to the `___init___.py` in both `functional` and `module`, as well as detailing a bit more the test example in `IzhikevichCell` and improving the parameters' names.

Jegp:
Nice. You didn't merge your changes with the upstream branch, though, but I took care of it for now.

When the checks pass, I'll merge it into the Norse repo branch so we can get Izhikevich models into Norse.

One final thing; did you consider implementing the other module types (SNNRecurrentCell, SNN, and SNNRecurrent)?

adelpierre:
I'm currently working on them, they should be faster to develop now that I have a better understanding of the code. I'll also see to write tests at the same time and a tutorial in parallel of my project.

Jegp:
> I'm currently working on them, they should be faster to develop now that I have a better understanding of the code. I'll also see to write tests at the same time and a tutorial in parallel of my project.

Brilliant! Ok, so if you fix the tests and add the modules we can proceed I think.

Jegp:
Nice work @adelpierre :-)

Issue #179, opened 2021-01-25
Implement IAF neuron


cpehle:
They are useful as baseline models and because certain digital neuromorphic chips like the ones by SynSense support them.

PR #204, opened 2021-04-12
IAF neuron initial implementation

cpehle:
Implements a simple integrate and fire neuron with pseudo-derivative gradient. Addresses #179.

### DIFF
--- a/norse/torch/functional/iaf.py
+++ b/norse/torch/functional/iaf.py
@@ -1,0 +1,79 @@
+import torch
+from typing import NamedTuple, Tuple
+from norse.torch.functional.threshold import threshold
+
+
+class IAFParameters(NamedTuple):
+    """Parametrization of a LIF neuron
+
+    Parameters:
+        v_th (torch.Tensor): threshold potential in mV
+        v_reset (torch.Tensor): reset potential in mV
+        method (str): method to determine the spike threshold
+                      (relevant for surrogate gradients)
+        alpha (float): hyper parameter to use in surrogate gradient computation
+    """
+
+    v_th: torch.Tensor = torch.as_tensor(1.0)
+    v_reset: torch.Tensor = torch.as_tensor(0.0)
+    method: str = "super"
+    alpha: float = torch.as_tensor(100.0)
+
+
+class IAFState(NamedTuple):
+    """State of a IAF neuron
+
+    Parameters:
+        z (torch.Tensor): recurrent spikes
+        v (torch.Tensor): membrane potential
+    """
+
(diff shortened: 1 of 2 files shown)


norse/torch/functional/iaf.py
```diff
@@ -0,0 +1,79 @@
...
+    """State of a feed forward IAF neuron
+
+    Parameters:
+        v (torch.Tensor): membrane potential
+    """
+
+    v: torch.Tensor
+
+
+def iaf_step(
```
Jegp, line 45: Documentation? Doesn't have to be a lot of text or equations

norse/torch/functional/iaf.py
```diff
@@ -0,0 +1,79 @@
...
+    v_new = (
+        v_new
+        + torch.nn.functional.linear(input_tensor, input_weights)
+        + torch.nn.functional.linear(state.z, recurrent_weights)
+    )
+
+    return z_new, IAFState(z_new, v_new)
+
+
+def iaf_feed_forward_step(
```
Jegp, line 66: A brief docstring would be helpful here as well

norse/torch/functional/test/test_iaf.py
```diff
@@ -0,0 +1,27 @@
...
+    iaf_feed_forward_step,
+)
+
+
+def test_iaf_feed_forward_step_batch():
+    x = torch.ones(2, 1)
+    s = IAFFeedForwardState(v=torch.zeros(2, 1))
+
+    z, s = iaf_feed_forward_step(x, s)
+    assert z.shape == (2, 1)
```
Jegp, line 16: I would recommend adding a one line test to validate that the spikes/state actually change over time. Mostly for regression purposes

norse/torch/functional/test/test_iaf.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_iaf_step_batch():
+    x = torch.ones(2, 10)
+    s = IAFState(z=torch.zeros(2, 5), v=torch.zeros(2, 5))
+    p = IAFParameters(v_reset=torch.zeros(2, 5))
+    w_in = torch.randn(5, 10)
+    w_rec = torch.randn(5, 5)
+
+    z, s = iaf_step(x, s, w_in, w_rec, p)
+    assert z.shape == (2, 5)
```
Jegp, line 27: Same as above

Jegp (approved):
Nice work. I'm happy to merge as-is, but it would be nice with docstrings. 
Modules would also be useful, but we can implement that later.

PR #206, opened 2021-04-25
Feature task

Jegp:
* Adds LSNN, LIF, and LSNN+LIF to memory task
* Fixes LSNN beta update


norse/dataset/memory.py
```diff
@@ -0,0 +1,100 @@
...
+            low=seq_periods // 2,
+            high=seq_periods,
+            size=(samples, seq_repetitions),
+        )
+
+    def __len__(self):
+        return self.samples
+
+    def _generate_sequence(self, idx, rep_idx):
+        data_pattern = torch.stack(
```
cpehle, line 60: I think the random number generator here needs to be seeded.

norse/torch/functional/lsnn.py
```diff
@@ -51,7 +50,7 @@ class LSNNParameters(NamedTuple):
 
     tau_syn_inv: torch.Tensor = torch.as_tensor(1.0 / 5e-3)
     tau_mem_inv: torch.Tensor = torch.as_tensor(1.0 / 1e-2)
-    tau_adapt_inv: torch.Tensor = torch.as_tensor(1.0 / 700)
+    tau_adapt_inv: torch.Tensor = torch.exp(torch.as_tensor(-1.0 / 800))
```
cpehle, line 53: where does this come from?

cpehle (approved):
LGTM, there are two places that are unclear to me.

norse/torch/functional/lsnn.py
Jegp: https://arxiv.org/pdf/1901.09049.pdf (26)
I liked the way they define the decay as proportional to the number of timesteps one would like to retain information. Meaning, that `exp^{-1 \over 800}` means that the decay roughly happens over 800 steps. This argument is taken from https://arxiv.org/pdf/1803.09574.pdf (2)

cpehle: Ok sure, but then you need to use ```exp(-dt/800)``` otherwise this doesn't work I believe.

Jegp:
Alright. This should also fix #182

PR #208, opened 2021-05-20
Preliminary support for sparse spike tensors in LIF Neurons

cpehle:


norse/torch/module/lif.py
```diff
@@ -51,13 +53,21 @@ class LIFCell(SNNCell):
...
     """
 
-    def __init__(self, p: LIFParameters = LIFParameters(), **kwargs):
-        super().__init__(
-            lif_feed_forward_step,
-            self.initial_state,
-            p=p,
-            **kwargs,
-        )
+    def __init__(self, p: LIFParameters = LIFParameters(), sparse=False, **kwargs):
```
Jegp, line 56: Excellent way to implement this.

norse/torch/module/lif.py
```diff
@@ -128,17 +138,29 @@ def __init__(
...
     ):
-        super().__init__(
-            activation=lif_step,
-            state_fallback=self.initial_state,
-            p=p,
-            input_size=input_size,
-            hidden_size=hidden_size,
-            **kwargs,
-        )
+        if sparse:
```
Jegp, line 145: At some point we should profile these kinds of checks. My intuition tells me it doesn't impact performance, but it'd be good to know. Fine for now, though.

Jegp (approved):
Excellent beginning. It's weird the output isn't sparse, which'll hinder the application for larger datapoints. But good start for now, so I'd say LGTM.
I'll add an issue to keep track of the implementation going forward. Ideally, we should either scale it to other models, or find a way that generalizes.

norse/torch/module/lif.py
cpehle: Yes, I agree.

PR #217, opened 2021-06-06
Fixed tensorboard import #215

### LINKED ISSUE #215: Tensorboard import error
When importing norse
```python
import norse.torch as snn
```

An import error over `tensorboard` occurs due to the tensorboard dependency in `norse/torch/utils/tensorboard.py`.

Solution: Use tensorboard through PyTorch `torch.utils.tensorboard`

Jegp:
Removed imports for `utils` on `norse` and `norse.torch` levels

This eliminates the need to install Tensorboard and fixes #215

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ Pipfile
 Pipfile.lock
 .pytype/
 htmlcov/
+.venv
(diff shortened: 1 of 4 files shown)


cpehle (approved):
Good catch.

PR #218, opened 2021-06-06
Tested application of sequentialstate

Jegp:
Simple test that we can actually apply `SequentialState` without errors


cpehle (approved):
LGTM

PR #219, opened 2021-06-06
Added support for nix

Jegp:
Adds support for Nix, where an environment can be spun up by executing `nix-shell` in the root folder


cpehle (commented):
Seems like a good idea 👍.

cpehle (approved):
LGTM

Issue #184, opened 2021-01-30
C++ modules are not loaded correctly on macOS


cpehle:
This code fails on macOS

```python
try:
    import torch
    from pathlib import Path

    print(Path(__file__).resolve().parent.parent)

    torch.ops.load_library(Path(__file__).resolve().parent.parent /  "norse_op.so"
    import sys

    setattr(sys.modules[__name__], "IS_OPS_LOADED", True)
except OSError:
    import sys

    setattr(sys.modules[__name__], "IS_OPS_LOADED", False)
```

cpehle:
Short term solution is to temporarily disable this optimisation and reintroduce it once the more comprehensive C++ primitives land.

Jegp:
As far as I understand it's only a problem in the test, right? Meaning, the actual code is unaffected. Can't we just survive the fact that they'll fail on mac until we get a chance to fix the bug?

cpehle:
> As far as I understand it's only a problem in the test, right? Meaning, the actual code is unaffected. 

Not entirely sure actually, in any case loading library like this is the wrong thing to do (TM). Because on windows/Mac the dynamic library postfixes are different. Separately the path the library is loaded from is also kind of wonky, not sure what to do about that.

PR #221, opened 2021-06-14
Feature cxx

Jegp:
* Re-introduced C++ as a fix for #184 
* Added cross-platform builds for Mac and Windows


.github/workflows/cpppackage.yml
```diff
@@ -0,0 +1,51 @@
+name: C++ package
+
+on: [push, pull_request]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Debug
```
cpehle, line 7: Shouldn't this be Release by default?

norse/torch/functional/test/test_lif.py
```diff
@@ -2,27 +2,28 @@
...
 
 import norse
-import norse.torch.functional.lif as lif_module
 from norse.torch.functional.lif import (
     LIFState,
     LIFFeedForwardState,
     LIFParameters,
     LIFParametersJIT,
     lif_step,
+    # lif_step_integral,
```
cpehle, line 11: comment?

cpehle (approved):
Great Job! Found only minor nitpicks.

.github/workflows/cpppackage.yml
Jegp: I've set tests to build when build type is `Debug`, so for the workflow this is correct.

norse/torch/functional/test/test_lif.py
Jegp: Good catch! I added the module, including some tests.

Jegp:
Thanks for the review @cpehle!

PR #230, opened 2021-07-24
Tested tensor datatype, fixes #228

### LINKED ISSUE #228: Force all voltage tensors to be floats
In https://github.com/norse/notebooks/issues/13 the voltages (copied from the input tensor) is of type `Long` and not `Float`, which crashes gradient calculations.

From my perspective, it makes most sense to retain event-based input tensors as numbers (`Int` or `Long`) and then ensure that the voltage is a floating point. Specifically, that would mean that we cast voltage and current tensors to floating points in, for instance, https://github.com/norse/norse/blob/master/norse/torch/functional/lif.py#L400

Curious to hear your input @biphasic

Jegp:
* Tests which datatype is fed into the lif modules
* Increases test coverage
* Refactors sparse code slightly
* Parallelizes tests

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -39,9 +39,9 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install -e .
-          pip install pytest tensorboard
+          pip install pytest-xdist tensorboard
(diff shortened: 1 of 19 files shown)


cpehle (approved):
LGTM

PR #231, opened 2021-07-26
Implement discrete adjoint sensitivity code

cpehle:
This is a first implementation of discrete adjoint sensitivity
equations. It has a number of drawbacks:
- The memory savings that are possible are not realised
- I did not carefully check the correctness
- It still uses BPTT and the fact that discrete Euler turns under BPTT into the discrete adjoint equations

### DIFF
--- a/norse/torch/module/lif.py
+++ b/norse/torch/module/lif.py
@@ -60,11 +60,16 @@ class LIFCell(SNNCell):
         dt (float): Time step to use. Defaults to 0.001.
     """
 
-    def __init__(self, p: LIFParameters = LIFParameters(), sparse=False, **kwargs):
+    def __init__(self, p: LIFParameters = LIFParameters(), **kwargs):
         super().__init__(
             activation=(
-                lif_feed_forward_step_sparse if sparse else lif_feed_forward_step
+                lif_feed_forward_adjoint_step
+                if p.method == "adjoint"
+                else lif_feed_forward_step
             ),
+            activation_sparse=lif_feed_forward_adjoint_step_sparse
+            if p.method == "adjoint"
+            else lif_feed_forward_step_sparse,
             state_fallback=self.initial_state,
             p=p,
             **kwargs,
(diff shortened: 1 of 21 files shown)


norse/torch/module/lif.py
```diff
@@ -53,14 +57,23 @@ class LIFCell(SNNCell):
         dt (float): Time step to use. Defaults to 0.001.
     """
 
-    def __init__(self, p: LIFParameters = LIFParameters(), sparse=False, **kwargs):
+    def __init__(
+        self, p: LIFParameters = LIFParameters(), sparse=False, adjoint=False, **kwargs
+    ):
         if sparse:
```
Jegp, line 63: I would suggest we merge this with #230, but perhaps into something similar to:
```python
activation = ...
super().__init__(activation=activation, ...)
```
That'll reduce a lot of redundant if-else blocks

Jegp (changes requested):
Theoretically, this looks great! Thanks for the contribution @cpehle 

Two things I think we *need* to look into are rebasing this with #230 and adding at least a single regression test for each modules that they at least run (can be as simple as just applying it with a zero tensor to make sure they don't crash).

Furthermore, I'd recommend we look a bit more into testing and fixing the constructor. The current approach to do `if .. elif .. else` for the module activation function seems error prone.

norse/torch/module/lif.py
Jegp: Another idea could be to wrap the modules instead of adding flags. The coexisting of the two flags indicate that it's possible to have both sparse *and* adjoint. We could perhaps do something like this:
```
class LIFCell(SNNCell):
    def__init__(self, activation=lif_feed_forward_step, ...):
        ...
class LIFCellAdjoint(LIFCell):
    def __init__(self):
        super().__init__(activation=lif_feed_forward_adjoint_step, ...)
class LIFCellSparse(LIFCell):
    def __init__(self):
         super().__init__(activation=lif_feed_forward_adjoint_step, ...)
```         
Alternatively, we can add an Enum or similar in the construction to ensure that the two flags aren't activated simultaneously.

Basically, I think we should think a bit about how we "toggle" the different activation modes of a module.

Jegp:
The current tests are failing because sparseness doesn't seem to be properly supported on MacOS.

I wouldn't feel entirely bad for leaving it out, but I know you're a Mac kinda guy @cpehle. Thoughts?

Issue #229, opened 2021-07-22
Implement Triangle Pseudo-Derivative


cpehle:
We should support the pseudo-derivative used in Bellec et al. 2018 and subsequent
work by other authors as implemented for example here: 

https://github.com/IGITUGraz/eligibility_propagation/blob/efd02e6879c01cda3fa9a7838e8e2fd08163c16e/Figure_3_and_S7_e_prop_tutorials/models.py#L40-L70

It has the advantage that it naturally lead to sparse gradients and therefore is a good alternative
to full blown event based gradient propagation.

Issue #207, opened 2021-05-20
Create a threshold module


cpehle:
Currently we handle threshold parametrisation by dispatching on a string.
This was originally done to support JIT specialisation. We should be able to accomplish the same by
using a torch.nn.Module which captures the necessary parameters of the chosen
threshold detection / pseudo-gradient computation.

Jegp:
What's exactly the purpose of this? Performance? Useability?

PR #232, opened 2021-07-31
Triangular surrogate gradient

Huizerd:
Solves #229. However, it seems that `HeaviTent` already implemented this surrogate gradient, so maybe I misunderstood the issue. I think this PR could be extended to address #207 as well.

I added some temporary visualization (`python norse/torch/functional/threshold.py`), and it seems several surrogate gradients are quite similar. Maybe this could be cleaned up as well?

Also, should we combine the `superspike.py` and `threshold.py` files into one?

### DIFF
--- a/norse/task/cifar10.py
+++ b/norse/task/cifar10.py
@@ -1,13 +1,8 @@
-import os
-import datetime
-import uuid
-
 from argparse import ArgumentParser
-from collections import namedtuple
 import torchvision
-import matplotlib.pyplot as plt
 
 import torch
+import torch.nn as nn
 import torch.nn.functional as F
 import torch.utils.data
 import pytorch_lightning as pl
@@ -17,8 +12,8 @@ from norse.torch.functional.lif import LIFParameters
 
 
 def label_smoothing_loss(y_hat, y, alpha=0.2):
-    log_probs = torch.nn.functional.log_softmax(y_hat, dim=1, _stacklevel=5)
-    xent = torch.nn.functional.nll_loss(log_probs, y, reduction="none")
+    log_probs = F.log_softmax(y_hat, dim=1, _stacklevel=5)
+    xent = F.nll_loss(log_probs, y, reduction="none")
     KL = -log_probs.mean(dim=1)
     loss = (1 - alpha) * xent + alpha * KL
     return loss.sum()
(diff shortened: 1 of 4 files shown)


cpehle:
You might be right and I simply forgot about the HeaviTent surrogate gradient. I think it would make sense to combine this with #207 and maybe also finally document the different options. I have some visualisations of the gradients lying around.

Jegp (changes requested):
Looks good. I would also kindly ask you to move the visualization code to the `utils` package. It's definitely useful.

Regarding the threshold module, I would be grateful to see that change in this PR. We can also easily move this to a later PR if you would like to just go ahead with a merge @Huizerd.

Thank you for your contribution! We look forward to many more ;-)

Huizerd:
Something like this?

```python
class ThresholdParameters(NamedTuple):
    """
    """

    method: Callable[[torch.Tensor, float], torch.Tensor] = superspike
    alpha: float = 100.0


class ThresholdFunction(torch.autograd.Function):
    """
    """

    @staticmethod
    @torch.jit.ignore
    def forward(ctx, x: torch.Tensor, p: ThresholdParameters) -> torch.Tensor:
        ctx.save_for_backward(x)
        ctx.p = p
        return heaviside(x)
    
    @staticmethod
    @torch.jit.ignore
    def backward(ctx, grad_output):
        (x,) = ctx.saved_tensors
        p = ctx.p
        grad_input = grad_output.clone()
        grad = p.method(x, p.alpha) * grad_input
        return grad, None


class Threshold(nn.Module):
    """
    """

    def __init__(self, p: ThresholdParameters = ThresholdParameters()):
        super().__init__()
        self.p = p
    
    def forward(self, x: torch.Tensor) -> torch.Tensor:
        return ThresholdFunction.apply(x, self.p)
```

And then additionally:
- a separate `surrogates.py` file with all the different surrogate gradients that can be used
- some nice documentation for each surrogate, and some sensible default parameters

Also:
- not sure how to handle multiple parameters for the surrogate; now there's only `alpha`, but some could use more (the triangular gradient in the works by Bellec can have 2, for instance)
- do you want to keep the `heaviside.py` as a separate file?
- the `heaviside` function has a `torch.jit.script` decorator, does this work when the calling function (e.g. `SuperSpike.forward` or `ThresholdFunction.forward`) has a `torch.jit.ignore` decorator?

Jegp:
I'd like to try to wrap this PR up, if possible. I know we've discussed parameterizing the surrogate derivatives as a module as well, but would it make sense to just push the triangle module as-is?

If that's ok with you, @Huizerd, could I ask you to remove the `###: temporary for visualization` statements so we can merge it?

Huizerd:
Yup sorry I forgot about this one a bit. Will adapt as you say, and work on the threshold module another time!

Huizerd:
Don't forget to unlink #207 :)

Jegp:
I've never seen this error before: https://github.com/norse/norse/pull/232/checks?check_run_id=3381810053#step:8:108
This doesn't seem like a relevant test. Would you mind adding this to the cifar file to avoid the failing type check?
```
# pytype: disable=attribute-error
... failing code
# pytype: enable=attribute-error
```

Further described here: https://google.github.io/pytype/user_guide.html

Jegp (approved):

Jegp:
Thanks a lot for sticking with us and contributing to the library @Huizerd! :fire:

PR #239, opened 2021-08-18
Add single step poisson encoding and singed poisson encoding

cpehle:
This is useful for adding PoissonEncodeStep and SignedPoissonStep to a SequentialState composition like so
```python 
model = SequentialState(
    SignedPoissonStep(), # will create a signed spike vector at each time step, provided the input is in range [-1,1]
    nn.Conv2d(1, 20, 5, 1),      # Convolve from 1 -> 20 channels
    LIFCell(),                   # Spiking activation layer
    nn.MaxPool2d(2, 2),
    nn.Conv2d(20, 50, 5, 1),     # Convolve from 20 -> 50 channels
    LIFCell(),
    nn.MaxPool2d(2, 2),
    nn.Flatten(),                # Flatten to 800 units
    nn.Linear(800, 10),
    LICell(),                    # Non-spiking integrator layer
)
```

### DIFF
--- a/norse/torch/functional/__init__.py
+++ b/norse/torch/functional/__init__.py
@@ -20,8 +20,10 @@ from .encode import (
     euclidean_distance,
     gaussian_rbf,
     poisson_encode,
+    poisson_encode_step,
     population_encode,
     signed_poisson_encode,
+    signed_poisson_encode_step,
     spike_latency_encode,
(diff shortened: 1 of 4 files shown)


Jegp:
LGTM

Jegp (approved):

PR #240, opened 2021-08-18
Implement lifted (in time) 2d convolution

cpehle:
This implements a way of efficiently performing a 2d convolution "lifted" / point-wise in time. This should bring performance 
benefits over using ```Lift(Conv2d)```, hence the name ```LConv2d```. Addresses #240.

### DIFF
--- a/norse/torch/module/conv.py
+++ b/norse/torch/module/conv.py
@@ -1,0 +1,80 @@
+import torch
+import torch.nn.functional as F
+from torch.nn.common_types import _size_2_t
+from torch.nn.modules.utils import _pair
+
+# pytype: disable=module-attr
+class LConv2d(torch.nn.Conv3d):
+    # pytype: enable=module-attr
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: _size_2_t,
+        stride: _size_2_t = 1,
+        padding: _size_2_t = 0,
+        dilation: _size_2_t = 1,
+        groups: int = 1,
+        device=None,
+        dtype=None,
+    ):
+        """Implements a 2d-convolution applied pointwise in time. See https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html?highlight=conv2d#torch.nn.Conv2d,
+        for documentation of the arguments, which we will reproduce in part here.
+
+        This module expects an additional temporal dimension in the tensor it is passed, that is
+        in the notation in the documentation referenced above, it turns in the simplest case a tensor
+        with input shape :math:`(T, N, C_{\text{in}}, H, W)` and output tensor of shape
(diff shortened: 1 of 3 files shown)


norse/torch/module/conv.py
```diff
@@ -0,0 +1,51 @@
+import torch
+import torch.nn.functional as F
+from torch.nn.common_types import _size_2_t
+from torch.nn.modules.utils import _pair
+
+
+class LConv2d(torch.nn.Conv3d):
```
Jegp, line 7: I would recommend calling it something more informed, like "LiftedConv2d" or "TimeConv2d".
Especially when there're no docs - which would also be nice ;-)

Jegp (changes requested):
I would suggest either that the name changes to something slightly more informative, or that we add heavy documentation to describe what's going on. Preferably both ;-)

norse/torch/module/conv.py
cpehle: Since this would be used a lot, I thought it would make sense to give it a short name.

Jegp: I never really understood the need for short variables. Most editors do code completion quite well. I agree we shouldn't call it ConvolutionInTwoDimensionsThatIsAlsoLiftedInTimeWhichIsCoolBecauseItIsFast :-)
Anyway, I don't have strong feelings, so if you think this is best, go ahead and GTM

Jegp (approved):

norse/torch/module/conv.py
cpehle: Well the idea is that you are then basically able to convert between a regular PyTorch network and a norse network by prepending a L. Basically I'm trying to make the following code nicer:
```python
model = Sequential(
    LConv2d(1, 20, 5, 1),      # Convolve from 1 -> 20 channels
    LIF(),                   # Spiking activation layer
    Lift(nn.MaxPool2d(2, 2)),
    LConv2d(20, 50, 5, 1),     # Convolve from 20 -> 50 channels
    LIF(),
    Lift(MaxPool2d(2, 2)),
    Lift( nn.Flatten()),                # Flatten to 800 units
    Lift(nn.Linear(800, 10)),
    LICell(),                    # Non-spiking integrator layer
)
```

Jegp: True, but that same logic could apply with "Lift". I do see your point. My main concern is towards newcomers who won't understand the "L". But, again, no strong feelings.

cpehle: Alternatively we could also explain that one can just use a Conv3d. But for that the shape ordering stands in the way.

cpehle: or we could omit the prefix alltogether, which would still be sort of confusing probably though

Jegp: Agreed, we should definitely distinguish it somehow. If you're happy with the code, let's just merge it as-is :)

PR #241, opened 2021-08-19
Updated pypi and conda build and bumped torch version requirement

### LINKED ISSUE #238: ERROR: Could not build wheels for norse which use PEP 517 and cannot be installed directly
**Describe the bug**
I get the following error when I install Norse from PyPi on a desktop PC (x64) with Windows 10: "ERROR: Could not build wheels for norse which use PEP 517 and cannot be installed directly". If I install it from Conda it installs v0.0.3 (noarch) instead of v0.0.6 (win-64).

I have tried this in Anaconda environments with Python 3.7, 3.8, and 3.9.

**To Reproduce**
From a Conda environment write: `pip install norse`

**Expected behavior**
You get this error: "ERROR: Could not build wheels for norse which use PEP 517 and cannot be installed directly".

**Desktop (please complete the following information):**
 - OS: Windows 10

Jegp:
Should fix #238 by using the https://github.com/pypa/build library to build sdist and wheels for windows and mac

In the meantime, I took the liberty to upgrade our manylinux build, so the Norse installation should be much faster going forward.

If you agree, I'll push out a 0.0.7RC1 release version to get this spinning for @lunzueta :)


cpehle (commented):
LGTM 👍.

cpehle (approved):

PR #242, opened 2021-08-20
Threshold module

Huizerd:
Attempts to solve #207. I started on implementing a base threshold module, which supports different kinds of surrogate gradients. However, then I noticed that some spiking functions were probabilistic, and I got stuck on trying to find a way to accomodate `forward()` functions with different arguments. Same goes a bit for the `backward()` ones: there might be some surrogates that have more parameters than only `alpha`.


Jegp:
Closing in favor of #307

PR #243, opened 2021-08-22
Changes to compile in Windows

lunzueta:
An attempt to fix the compilation of Norse in native Windows. This works locally in my machine. It seems that I can't test the Github workflow of this by myself. Let's do it together. My main doubt is whether the Visual Studio building setup will be correct or not in that workflow machine (you'll see that I execute a VS-related .bat file with a specific path, from the .bat file I added). I see that there are people ([example](https://gist.github.com/NickNaso/0d478f1481686d5bcc868cac06620a60)) that apply the Visual Studio building setup (for C++) in Github workflow, so, in principle, it should be possible.


set_win_env.bat
```diff
@@ -0,0 +1,2 @@
+set DISTUTILS_USE_SDK=1
+"C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
```
Jegp, line 2: Is this the same path for all users? 

Another thought; would it make sense to use something like https://chocolatey.org/ to make sure the correct packages are installed?

Jegp:
Nice to-the-point PR, thank you!

I had one thought about dependency/package management that I'm curious to hear your opinion about. Afterwards, I'll try to merge this and see what happens when we try to push out another release.

Jegp (approved):

set_win_env.bat
lunzueta: No, that's why I'm not totally happy with my "fix" and I said in the other thread that it wasn't "elegant". In Windows, you might have the Visual Studio IDE installed (with its building tools) or just the Visual Studio Build Tools (as I do, and for the IDE I use VSCode). That's the path of the environment batch file from VS2019 Build Tools. The path to that file could also be for example "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat" if you have VS2019 Enterprise edition installed, or another with other VS versions and editions (yep, it's a total pain, hehe). I see that in Chocolatey you have [this](https://community.chocolatey.org/packages/visualstudio2019buildtools), so it could be a good solution for this. I don't know how to use it but I might investigate a little bit :-)

lunzueta: I've edited the batch file to cover all the possibilities with VS2019. Maybe we might test this first on Github Workflow as it is and see if it works (not sure what the Windows remote machine has installed), and if so, document it for the Windows installation. And later we might try to do something with Chocolatey? I say this because I don't fully like forcing users to install VS Build Tools if they already have VS2019 installed, it could be a bit redundant for some users. Besides, newer VS versions will be coming. Not easy to solve this Windows thing...

lunzueta: Well, I have gone a little bit deeper with Chocolatey. It's easy to install VS2019 Build Tools once you have it installed, in a similar way to `apt`. The bad part is that you need be admin to install Chocolatey from the command prompt. Would that be possible?

Jegp: I think it's possible in the containers here, but I don't know for sure. And, as we said, it's hard to test :-)

Is it correct that the `vcvars` script initializes settings required for compilation on WIndows? Would it then make sense to use a workflow like this to setup the vars then? https://github.com/marketplace/actions/visual-studio-shell Their script seems to do something similar but less dependent on specific paths: https://github.com/egor-tensin/vs-shell/blob/master/action.yml

lunzueta: Yes, That's it. In the end, it's only to set DISTUTILS_USE_SDK=1 and then initialize vcvars (apart from putting the additional argument I included in setup.py for C++17, which is also required in this case), and compile normally.

Jegp: Alright. Let's try this. It should just work for the pip build, so I'm not too concerned about spilling variables. Thanks again for the contribution. I hope I can count on you to have a second try if it doesn't work ;-)

cpehle (commented):
If I recall correctly using (vcvars.bat)  manually can be avoided by using the appropriate developer command prompt. This solution would not work for example for a different version of Microsoft Visual Studio.

Jegp:
Sure. This only needs to work for the build workflow. Is there a quick solution we can apply for the GitHub actions you would prefer @cpehle ?

setup.py
```diff
@@ -16,7 +16,7 @@
 cpp_extension = CppExtension(
     name="norse_op",
     sources=["norse/csrc/op.cpp", "norse/csrc/super.cpp"],
-    extra_compile_args=["-O3"],
+    extra_compile_args=["-O3", "/std:c++17"],
```
cpehle, line 19: this wouldn't work on Linux.

cpehle (changes requested):

setup.py
lunzueta: OK. I've added one more change to solve this. I hope you like it.

set_win_env.bat
lunzueta: No problem. Thank you guys for sharing this great software with the community.

cpehle (approved):

cpehle:
> Sure. This only needs to work for the build workflow. Is there a quick solution we can apply for the GitHub actions you would prefer @cpehle ?

I'm not sure, I thought this was already set in the windows-latest build environment to be honest.

Jegp:
> > Sure. This only needs to work for the build workflow. Is there a quick solution we can apply for the GitHub actions you would prefer @cpehle ?
> 
> I'm not sure, I thought this was already set in the windows-latest build environment to be honest.

Yep, I had the same impression. I'm basically happy that @lunzueta is happy to help us out :-) This seems to work since all the builds are passing, so I'd suggest we merge this and try an RC release. We wanted to have 0.0.7 out for ages anyway.

PR #251, opened 2021-08-31
Include MNIST training example + modify learning notebook

cpehle:


Jegp (changes requested):
Great to have an MNIST example. And the learning page is much more illustrative than before. I think we should consider two things:

- I appreciate removing the factual problems in the `learning` page, but I do think the current version is lacking some of the intuitive explanation I tried to provide before. The current page provides no introduction to spiking neurons, no explanation why we need surrogate gradients, and, in my opinion, not enough practical information on how to train SNNs with Norse. I think the details are great, but I would perhaps suggest keeping the (admittedly, really nice) visualization in a separate page about surrogate gradients. I would suggest the 'learning' page is a quick tour about how surrogate gradients work and how it's used in Norse.
- There are red warnings/errors in the MNIST dataset and it doesn't compile. How about only using a limited dataset and set epoch=1 as default to avoid build-time slowdown? We can always ask people to augment the data/epochs if they need. I doubt that'll interfere with the purpose of the notebook.

Jegp:
Should this be closed, btw @cpehle ?

cpehle:
I think this is still worth working on, the current documentation on learning is not consistent with my view of the role of surrogate gradients. We should try rigorously explain what the challenges in defining gradient based learning in spiking neural networks are, this is not really done by the current doc at the moment.

Issue #250, opened 2021-08-30
Spike-time decoding


Jegp:
Add a function and module that permits spike-time decoding, as suggested by @schmitts https://twitter.com/sbstnschmtthd/status/1432343373072019461

schmitts:
Thanks. I'll be happy to test anything.

Jegp:
You're more than welcome to test out the implementation and provide feedback. I'm open to suggestions on how this can be improved/built upon.

cpehle:
This is addressed by #321 , @schmitts 🎉, so feel free to give it a try 🙂.

PR #255, opened 2021-09-08
Added a spike time decoder

Jegp:
This decoder addresses #250 by computing the indices of the spike values (1's) in a given tensor. 

The returned tensor has shape (`ndim`, `nvalues`), e. g.

```python
x = torch.tensor([0, 0, 0, 1, 0])
SpikeTimeDecoder()(x)           
    # tensor([[3]])
```

```python
import norse.torch as nt
model = nt.SequentialState(
    LIFCell(),
    nt.SpikeTimeDecoder()
)
x, _ = model(torch.ones(10, 2))
   # tensor([[6, 6, 9, 9]
   #         [0, 1, 0, 1]])```

### DIFF
--- a/norse/torch/functional/__init__.py
+++ b/norse/torch/functional/__init__.py
@@ -31,6 +31,7 @@ from .correlation_sensor import (
     correlation_based_update,
     correlation_sensor_step,
 )
+from .decode import sum_decode, spike_time_decode
 from .encode import (
     constant_current_lif_encode,
     euclidean_distance,
@@ -171,6 +172,9 @@ __all__ = [
     "CorrelationSensorState",
     "correlation_based_update",
     "correlation_sensor_step",
+    # Decoders
+    "sum_decode",
+    "spike_time_decode",
     # Encoders
     "constant_current_lif_encode",
     "euclidean_distance",
--- a/norse/torch/functional/decode.py
+++ b/norse/torch/functional/decode.py
@@ -17,6 +17,32 @@ def sum_decode(tensor: torch.Tensor, dimension: int = 0):
     return tensor.sum(dimension)
 
 
+class SpikeTimeDecode(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, i: torch.Tensor):
+        if i.is_sparse:
+            sparse = i.coalesce()
+        else:
+            sparse = i.to_sparse().coalesce()
+        ctx.save_for_backward(sparse, torch.as_tensor(i.is_sparse))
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by Jegp** [comment]
It just occurred to me that it may be interesting to not get the time *indices*, but the actual *time* of the simulation. We would achieve that by adding the `dt` as a parameter:
```python
def spike_time_decode(x, dt: float=0.001):
    ...
```

cpehle (changes requested):
This doesn't support gradient propagation, I believe. I think a ```SpikeTimeDecoder``` should support propagation of gradients, since otherwise it can't be used in training.

Jegp:
I must admit I have a hard time wrapping my head around this. The decoder returns time indices from the spikes. If any arithmetic is being done on the timesteps (say, finding ISIs), how would those gradients flow from the indices to the spikes themselves?

cpehle:
> I must admit I have a hard time wrapping my head around this. The decoder returns time indices from the spikes. If any arithmetic is being done on the timesteps (say, finding ISIs), how would those gradients flow from the indices to the spikes themselves?

Well this would need to be accomplished by implementing the decoder as a custom torch.Function. This still doesn't give you exactly what one wants, since the spike times are tied to the integration grid, but it would allow one to correctly transition between a differentiable function of the spike times (with the limitation being that the spike times itself are actually on a grid) and the hybrid system that produced them.

Jegp:
Alright. After some consideration, I added gradients that basically pools the gradient values for each index values (shaped `(ndim, nvalues)`) over the 1st dimension.

Note that the returned indices are now floats to support the flow of gradients.

Jegp:
Not sure if this is the perfect way of doing it, though. Open for inputs!

Jegp:
Unless I hear anything, I may just merge this as-is. We can always improve it later, but then it's at least available.

cpehle:
> Unless I hear anything, I may just merge this as-is. We can always improve it later, but then it's at least available.

I think at a minimum this should come with a working example, where the spike time decoder actually is used in a task. Otherwise we can't know for sure, whether the current version is actually useful. I don't think we should be adding features, which we are unsure about, whether they actually can be used in a practical setting.

Jegp:
Superseded by #321

cpehle:
It might make sense to keep this as an alternative to #321, which we can test once we have a task implemented that uses (first) spike time decoding and after splitting the implementation into a module and functional part. We still need to check that this way of assigning errors, produces sensible results.

PR #256, opened 2021-09-16
Update sequential.py

AlmaLuna94:
Changed 1st example in SequentialState. Second example still doesn't work though.
These examples are also on the webpage https://norse.github.io/norse/generated/norse.torch.module.SequentialState.html, In as of yet unchanged form.


Jegp:
Thanks for the contribution! I'll validate that the checks pass and merge asap :-)

Jegp:
Thanks again. If you have time to resolve the second example, I would also be grateful :)

PR #257, opened 2021-09-16
Expand introduction to spiking systems

cpehle:
I did write some more text on the spiking systems doc, hopefully more accessible than last time. I think we should stay away from using "activation" though, because SNN are more like recurrent neural networks.


Jegp (approved):
The level of detail and description here is excellent. This really hits home for the new and "uninitiated" users of SNNs and is, to me, a vast improvement compared to the last iteration. I think the current version is much better and can be merged as-is. But I do have some suggestions for improvements.

* The link to the leaky integrator doesn't work for me (`Unknown type`)
* As per the agenda in the top of the notebook you are indeed simulating and visualizing neurons, but the notebook does not seem to do actual optimization. Would it make sense to either add the optimization part or simple leave it out? For now. We can always add it later.

PR #266, opened 2021-10-07
Update Readme

cpehle:
We don't implement the Superspike method, only the SuperSpike surrogate gradient function. This came up in a Zoom call today, so I think it makes sense to clarify this.


Jegp (approved):
LGTM! :)

PR #267, opened 2021-10-14
Add spack package file and run spack build test

muffgaga:
This adds a spack package file that successfully built on our local spack instance… however, this should be probably reflected in your github workflow → should I just try to add it (i.e. don't merge this but I'll update this PR) :)?

### DIFF
--- a/.github/spack/build_spack_package.sh
+++ b/.github/spack/build_spack_package.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -euxo pipefail
+shopt -s inherit_errexit
+
+WORKSPACE=${PWD}
+
+TMP_DIR=$(mktemp -d)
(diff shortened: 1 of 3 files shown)


muffgaga:
@Jegp 

> Looking at the build checks, does it really take >2h to build the spack package?! I assume this will improve with caching?

Spack is a build-from-source package manager… in this case it includes a full software stack with CPython and PyTorch (but excluding gcc/binutils which it takes from the host OS). The CI VMs are also rather small…
And yes, we can use a build cache to skip all builds except for the norse build → that's probably a valid strategy for norse PRs (as you probably don't want to test changes to the underlying software there but rather the changes themselves).

To verify that norse still works with other spack packages a test against spack HEAD would probably be still valuable.

PR #271, opened 2021-11-05
Updated benchmarks and added GeNN optimizations

Jegp:


Jegp:
We've been beaten! GeNN added some form of caching mechanism that dramatically speeds up their runtime.

cpehle (changes requested):
I don't think the changes to the adjoint code should be mixed with benchmark code changes.

Jegp:
> I don't think the changes to the adjoint code should be mixed with benchmark code changes.

Agreed!

norse/benchmark/norse_lif.py
```diff
@@ -29,7 +32,7 @@ def forward(
 
         for ts in range(sequence_length):
             x = self.fc(input_spikes[ts])
-            z, s = _lif_feed_forward_step_jit(input_tensor=x, state=s, p=p, dt=self.dt)
```
cpehle, line 32: why this change?

norse/torch/functional/lif.py
```diff
@@ -465,13 +413,28 @@ def lif_feed_forward_step_sparse(
     i_decayed = state.i + di
 
     # compute new spikes
-    z_new = threshold(v_decayed - p.v_th, p.method, p.alpha)
```
cpehle, line 468: this shouldn't be mixed in here either ...

norse/torch/functional/lif.py
```diff
@@ -465,13 +413,28 @@ def lif_feed_forward_step_sparse(
...
+    z_new = torch.sparse_coo_tensor(
+        indices=thresholds.indices(),
+        values=jumps,
+        size=v_decayed.size(),
+        device=thresholds.device,
+    ).coalesce()
+    # z_new = threshold((v_decayed - p.v_th).to_dense(), p.method, p.alpha)
+    # z_new = z_new.to_sparse().coalesce()
     # compute reset
-    v_new = (1 - z_new) * v_decayed + z_new * p.v_reset
```
cpehle, line 470: nor this

cpehle (changes requested):
I found some additional issues.

cpehle:
Also does the updated code for GeNN run 32 instances of the experiment (batch size = 32)?

norse/benchmark/norse_lif.py
Jegp: I took the liberty to remove the checks for C++ availability in the functional API. Here, I'm directly calling the compiled C++ version. Would you suggest a different approach?

norse/torch/functional/lif.py
Jegp: Probably not :)

Jegp:
> Also does the updated code for GeNN run 32 instances of the experiment (batch size = 32)?

Yes: https://github.com/norse/norse/blob/feature-benchmark/norse/benchmark/genn_lif.py#L40

Jegp:
Alright, I finally got the chance to play with CUDA streams, and it's working neatly:



The JIT functions *appear* to be more performant than the native C++ code I've written. We should probably consider moving to JIT entirely...

Jegp:
Also, GeNN is parallelizing on the batch dimension. Meaning, they're spawning parallel CUDA kernels for each batch entry. We can compete on that, but I think that kind of defeats the purpose (it's a lot of coding).

Either we
* Leave it as is and merge the PR
* Migrate GeNN to use `batch * feature` population sizes *once*, with the argument that we're measuring a single tensor flowing, not the parallelization of multiple graphs.

In any case, it'd be helpful to have more complex benchmark for larger networks.

cpehle:
> Also, GeNN is parallelizing on the batch dimension. Meaning, they're spawning parallel CUDA kernels for each batch entry. We can compete on that, but I think that kind of defeats the purpose (it's a lot of coding).
> 
> Either we
> 
> * Leave it as is and merge the PR
> * Migrate GeNN to use `batch * feature` population sizes _once_, with the argument that we're measuring a single tensor flowing, not the parallelization of multiple graphs.
> 
> In any case, it'd be helpful to have more complex benchmark for larger networks.

What I don't understand is: Does GeNN actually simulate batch size different instances of the network with distinct poisson input and different parameters or not?

Jegp:
> What I don't understand is: Does GeNN actually simulate batch size different instances of the network with distinct poisson input and different parameters or not?

That's what I can understand from this: https://github.com/norse/norse/blob/feature-benchmark/norse/benchmark/genn_lif.py#L40
It looks like they, for every batch, create independent populations for the Poisson input, linearities, and LIF models. Which allows them to simulate each batch independently in parallel. That's what I mean with 'defeating the purpose' because the model is split into `batch_size` number of parallel models, which is of course more performant, but also a tremendous amount of work if that is to be done for each layer. On the other hand, that's exactly how CUDA kernels should be used if performance is all that matters. Honestly, I'm torn. But perhaps it makes sense to leave them the benefit of the doubt. I don't think the numbers they get is wrong in that sense.

neworderofjamie:
Just FYI, this is also a sub-optimal (and ugly) way of simulating parallel batches in GeNN. You can just set ``model.batch_size = X``. Happy to do that if you want a better comparison.

Jegp:
Thanks for the input Jamie! The goal for me is really accuracy and fairness. If the changes you suggest are more fair that way, I would be more than happy if you could add them in. 

I would also love to have benchmarks on a model with a few layers. Would it make sense for me to sketch a model in Norse that you can implement in GeNN?

neworderofjamie:
Sure - I'll have a go this afternoon. In the norse version, are the same weights shared between each batch or are they different?

Also, when running the benchmark, I get these errors after every line of output:
```
/usr/lib/python3.8/multiprocessing/resource_tracker.py:203: UserWarning: resource_tracker: There appear to be 1 leaked shared_memory objects to clean up at shutdown
```
any thoughts?

cpehle:
> Sure - I'll have a go this afternoon. In the norse version, are the same weights shared between each batch or are they different?

Weights are shared, thanks for looking into this @neworderofjamie!

neworderofjamie:
Done! Feel free to merge in https://github.com/neworderofjamie/norse/tree/genn_example_fix

Jegp:
Thansk a lot! I'll look at them soon - and look into creating a larger benchmark for a more interesting model. Probably in a different PR though :-)

neworderofjamie:
Cool! Would be happy to implement a bigger model with some sparse connections or convolutionals or something

cpehle (approved):

cpehle:
We should fix the formatting issue, otherwise I think this is good to go 👍, thanks to @neworderofjamie aswell for the effort :).

Jegp:
New graph. Looks roughly identical to the previous one.

PR #277, opened 2021-12-05
Add LifRefracRecurrent layer

Pugavkomm:
Added layer LIFRefracRecurrnet. I tried to do it in the style of the entire library code.


Pugavkomm:
I tried to add a LIFRefractRecurrent layer, but I still don't really understand the reason for the error that comes out during testing. I also ask you to look at the code and point out possible errors that I might not have noticed in the implementation. Especially, you need to check  `backward` in class `LIFSparseAdjointRefractFunction`.

norse/torch/functional/adjoint/lif_refrac_adjoint.py
```diff
@@ -112,6 +113,122 @@ def lif_refrac_adjoint_step(
     return z, LIFRefracState(LIFState(z, v, i), rho)
 
 
+class LIFSparseAdjointRefractFunction(torch.autograd.Function):
```
Jegp, line 116: I would recommend choosing a consistent abbreviation for 'refractory'. Below, `Refrac` is used but here it's `Refract`.

norse/torch/functional/adjoint/lif_refrac_adjoint.py
```diff
@@ -112,6 +113,122 @@ def lif_refrac_adjoint_step(
...
+def lif_refrac_adjoint_step_sparse(
+    intput: torch.Tensor,
+    s: LIFRefracState,
+    input_weights: torch.Tensor,
+    recurrent_weights: torch.Tensor,
+    p: LIFRefracParameters,
+    dt: float = 0.001,
+) -> Tuple[torch.Tensor, LIFRefracState]:
+    z, v, i, rho = LIFSparseAdjointRefractFunction.apply(
+        input, s.lif, s.lif, s.lif, s.rho, input_weights, recurrent_weights, p, dt
```
Jegp, line 226: The `s.lif, s.lif, s.lif` arguments look strange. Shouldn't it be a single `LIFState` instance, as per `LIFSparseAdjointRefractFunction::forward`? This could also be where your type error comes from.

Jegp (changes requested):
Nice work overall. I would maybe look into the typing error in the forward pass of the `LIFSparseAdjointRefractFunction` function. 
Also, if you wouldn't mind adding a single unit test for the *functional* part of the LIFRefrac part. Then we have regression tests for future references :-)

norse/torch/functional/adjoint/lif_refrac_adjoint.py
Pugavkomm: I have added some tests. Which test should be added yet

norse/torch/functional/adjoint/lif_refrac_adjoint.py
Pugavkomm: Oh, thank you. Obviously a mistake here, for some reason I did not pay attention when I looked through it many times

norse/torch/functional/adjoint/lif_refrac_adjoint.py
Jegp: Thanks for the changes. I was thinking about tests for the functional parts, but that's already been done I see!

norse/torch/functional/adjoint/lif_refrac_adjoint.py
Pugavkomm: I fixed it but it didn't fix the error.

Jegp: I think the `self` parameter in `lif_refrac` was "pushing" the other parameters down, so the type checker interpreted the refrac parameters as the dt. Let's see if the checks agree!

Jegp (approved):
Thanks for seeing this through! I'm pretty happy about it as-is. There are some documentation improvements we probably should do one day, but it should be functional for now :-)

PR #281, opened 2021-12-19
Added IAF and box modules

Jegp:
Added a feedforward module for the integrate-and-fire functionals.

Added functionals and modules for LIF "box" models without the current term. Suggestions for better names are welcome! :-)


cpehle (approved):
LGTM

PR #291, opened 2022-01-13
Feature lif adex recurrent refrac

Pugavkomm:


norse/torch/module/__init__.py
```diff
@@ -78,6 +86,13 @@
 )
 
 __all__ = [
+    # LIFAdExRafrac
```
Jegp, line 89: Can this be removed?

norse/torch/functional/test/test_lif_adex_refrac.py
```diff
@@ -0,0 +1,36 @@
...
+    s = LIFAdExRefracState(
+        lif_adex=LIFAdExState(
+            z=torch.zeros(10), v=torch.zeros(10), i=torch.zeros(10), a=torch.zeros(10)
+        ),
+        rho=torch.zeros(10),
+    )
+    input_weights = torch.randn(10, 20).float()
+    recurrent_weights = torch.randn(10, 10).float()
+    for _ in range(100):
+        _, s = lif_adex_refrac_step(x, s, input_weights, recurrent_weights)
```
Jegp, line 24: I appreciate that you're testing this. I would perhaps suggest adding a test for the numerical output, just so there is some checks against the correctness of the calculations.

norse/torch/functional/test/test_lif_adex_refrac.py
```diff
@@ -0,0 +1,36 @@
...
+def test_lif_refrac_feed_forward_step():
+    x = torch.ones(10)
+    s = LIFAdExRefracFeedForwardState(
+        lif_adex=LIFAdExFeedForwardState(
+            v=torch.zeros(10), i=torch.zeros(10), a=torch.zeros(10)
+        ),
+        rho=torch.zeros(10),
+    )
+    for _ in range(100):
+        _, s = lif_adex_refrac_feed_forward_step(x, s)
```
Jegp, line 36: Same as above. Could you add some numerical validation?

Jegp (changes requested):
Thanks for the addition! I hope you understand that my comments about the testing relate to regression/completeness further down the line. If we modify something in the future, it's helpful to know that we don't mess up the neuron logic :-)

cpehle (approved):
I think this is pretty good as is. We should write numerical tests of the AdEX equations ideally against other simulators, still. But this goes beyond just asserting the values that the integration happens to produce, but instead we could look into reproducing the Naud patterns for example. But this is for another submission, I think.

Jegp:
Thank you for merging this @cpehle ! And sorry for dragging my feet @Pugavkomm

PR #301, opened 2022-02-28
feature: include ying-yang dataset

cpehle:
Adds the Ying-Yang dataset to the norse code base. Intention is to quickly be able to demonstrate time-to-first spike learning with the Norse library.

### DIFF
--- a/norse/dataset/test/test_ying_yang.py
+++ b/norse/dataset/test/test_ying_yang.py
@@ -1,0 +1,6 @@
+from norse.dataset.yingyang import YinYangDataset
+
+def test_generation():
+    dataset_train = YinYangDataset(size=20, seed=42)
+    dataset_validation = YinYangDataset(size=30, seed=41)
+    dataset_test = YinYangDataset(size=40, seed=40)
--- a/norse/dataset/yingyang/__init__.py
+++ b/norse/dataset/yingyang/__init__.py
@@ -1,0 +1,118 @@
+# MIT License
+# 
+# Copyright (c) 2020 lkriener
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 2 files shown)


norse/dataset/test/test_ying_yang.py
```diff
@@ -0,0 +1,6 @@
+from norse.dataset.yingyang import YinYangDataset
+
+def test_generation():
+    dataset_train = YinYangDataset(size=20, seed=42)
+    dataset_validation = YinYangDataset(size=30, seed=41)
+    dataset_test = YinYangDataset(size=40, seed=40)
```
Jegp, line 6: Perhaps add an assert on the shape, just for posterity?

norse/dataset/yingyang/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+Initializing the dataset:
+
+.. code: python
+    from norse.dataset.yingyang import YinYangDataset
+
+    dataset_train = YinYangDataset(size=5000, seed=42)
+    dataset_validation = YinYangDataset(size=1000, seed=41)
+    dataset_test = YinYangDataset(size=1000, seed=40)
+
+**Note** It is very important to give different seeds for trainings-, validation- and test set, as the data is generated randomly using rejection sampling. Therefore giving the same seed value will result in having the same samples in the different datasets!
```
Jegp, line 41: An approach to the random seed problem I've seen several times would be to leave the seed unset (default param `None`) to default to random values. In that way ignorant/busy users won't accidentally reuse the seed.
I think I would prefer that approach to trusting users to read the docs (yeah, right) and correctly set the parameter.

Jegp (changes requested):
Neat addition, good docs.

I would like to address/discuss the random seed issue, otherwise we're good to go!

norse/dataset/yingyang/__init__.py
cpehle: Yeah that makes sense :).

Jegp:
I'll suggest adding this to the [Tonic library](https://tonic.readthedocs.io/)

PR #304, opened 2022-03-23
Add some interactivity to the introduction to spiking networks + plot style

cpehle:


Jegp:
These changes look great! LGTM.

However, I can't get the current binder/Thebes integration working in the browser. It'd be useful to investigate this so, users can exploit the interactivity directly in the browser.

Jegp (approved):

PR #305, opened 2022-03-23
feature: investigate norse module jitting

cpehle:
This is the beginning of a longer set of features that we should tackle both to improve performance and make integration with hardware backends easier. TorchScript has certain restrictions for which kinds of constructs it allows and ideally all core norse modules should be compatible with it. This could mean that some of the more "convenient" features we have implemented might need to be rethought.

The first commit just tests out the minimal set of modifications necessary to make jitting a sequence of norse modules work using the example of the LIF neuron model.


Jegp:
I would say this is related to #389 where we're testing for `torch.compile`.

PR #307, opened 2022-03-24
Threshold enum

cpehle:
This change introduces an enum for the possible surrogate gradient functions we support. This is an improvement over the string based dispatch we did before (due to limitations of the jit), but we might want to use a torch.Module instead (#242).

It also undoes hardcoding the super spike surrogate gradient method for `lif` neurons, which shouldn't be there in the first place.


norse/torch/functional/decode.py
```diff
@@ -15,3 +15,23 @@ def sum_decode(tensor: torch.Tensor, dimension: int = 0):
     Sums the input tensor in the first dimension by default.
     """
     return tensor.sum(dimension)
+
+
+class FirstSpikeTimeDecode(torch.autograd.Function):
```
Jegp, line 20: This should be merged with #255

Jegp (changes requested):
I think this is a good step forward, and we should merge it after the spike time decoder have been moved to #255.

I would like to discuss an actual `Threshold` module inherited from `torch.nn.Module`, as begun by @Huizerd in #242. I think that's a cleaner and more maintainable approach. Roughly, the code would look as follows:

```python
class SurrogateThreshold(torch.nn.Module):
   ...

class SuperSpikeThreshold(SurrogateThreshold):
   ...

def lif_...(x, p: Parameters = LIFParameters(threshold=SuperSpikeThreshold()):
   z = threshold(x)
```

This would get rid of the weird extra `alpha` parameter, because it would all be parameterized in the module.

Jegp:
Closing as stale. Happy to revive if needed

PR #310, opened 2022-04-07
Corrected adex docs

Jegp:


spiking-transformer (approved):
LGTM, maybe we should add a caveat regarding Euler integration precision as well.

PR #321, opened 2022-05-09
Feature spikes to times decoder

lucablessing:
Class to decode spike input, with time as first dimension, into times / indices of spikes and custom autograd function.

### DIFF
--- a/norse/torch/functional/spikes_to_times_decoder.py
+++ b/norse/torch/functional/spikes_to_times_decoder.py
@@ -22,7 +22,6 @@ class SpikesToTimesDecoder(torch.nn.Module):
         :param spike_count: number of elements (ascending) to return for each output channel.
         :param convert_indices_to_times: Whether to return times or indices along input's time axis
         :param dt: time step of input's time axis (needed for conversion of indices to times)
-        :param device: cpu, gpu etc.
         """
         super().__init__()
         self.spike_count = spike_count
@@ -36,7 +35,7 @@ class SpikesToTimesDecoder(torch.nn.Module):
 
         :param spike_input: spike tensor with 0s and 1s (indicating spikes)
         """
-        spike_indices = self.decoding_fn(spike_input, self.spike_count, self.device)
+        spike_indices = self.decoding_fn(spike_input, self.spike_count)
         if self.convert_to_time:
             spike_indices = spike_indices * self.dt
         return spike_indices
@@ -50,7 +49,7 @@ class SpikesToTimesDecoder(torch.nn.Module):
         """
 
         @staticmethod
-        def forward(ctx, spike_input, spike_count, device):
+        def forward(ctx, spike_input, spike_count):
             """
             Return indices of first spike_count spikes (if spike_count < spike_input.shape[0],
             i.e. its time dim). If less than spike_count spikes happened along a output trace,
@@ -96,4 +95,4 @@ class SpikesToTimesDecoder(torch.nn.Module):
 
             grad_input = -1.0 * grad_output * grad_local
 
-            return grad_input, None, None
+            return grad_input, None
--- a/norse/torch/functional/spikes_to_times_decoder.py
+++ b/norse/torch/functional/spikes_to_times_decoder.py
@@ -1,0 +1,99 @@
+import torch
+
+
+class SpikesToTimesDecoder(torch.nn.Module):
+    """
+    Module wrapper for spike to spike-time (or spike-index) decoder.
+
+    Decodes input spikes with shape [time_steps x batch_size x output_size] into spike indices
+    tensor with shape [spike_count x batch_size x output_size] in ascending order, e.g. for
+    spike_count = 1 returns only times of first spikes along each batch- and output-dimension.
+    If stated, converts indices along time axis into times with given dt value.
+    """
+
+    def __init__(
+        self,
+        spike_count: int = 1,
+        convert_indices_to_times: bool = True,
+        dt: float = 1e-3,
+    ):
+        """
+        Initialize decoder.
+        :param spike_count: number of elements (ascending) to return for each output channel.
+        :param convert_indices_to_times: Whether to return times or indices along input's time axis
+        :param dt: time step of input's time axis (needed for conversion of indices to times)
+        :param device: cpu, gpu etc.
+        """
+        super().__init__()
+        self.spike_count = spike_count
+        self.convert_to_time = convert_indices_to_times
+        self.dt = dt
+        self.decoding_fn = self.ToSpikeTimes.apply
+
+    def forward(self, spike_input):
+        """
+        Call decoder custom autograd function
+
+        :param spike_input: spike tensor with 0s and 1s (indicating spikes)
+        """
+        spike_indices = self.decoding_fn(spike_input, self.spike_count, self.device)
+        if self.convert_to_time:
+            spike_indices = spike_indices * self.dt
+        return spike_indices
+
+    @staticmethod
+    class ToSpikeTimes(torch.autograd.Function):
+        """
+        Convert spike input with shape [time_steps x batch_size x output_size] to the indices of
+        spike times, shape [spike_count x batch_size x output_size]. If no spike is present, time
+        or index is set to inf.
+        """
+
+        @staticmethod
+        def forward(ctx, spike_input, spike_count, device):
+            """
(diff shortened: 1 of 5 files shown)


cpehle (changes requested):
Thanks for this contribution! Could you please move the Module implementation into norse/torch/module and leave the autograd function as is. This is a borderline case but I still would think it makes sense to do so.

lucablessing:
> Could you please move the Module implementation into norse/torch/module and leave the autograd function as is.

Done. The changes hopefully should be like you requested.

cpehle (approved):
LGTM, but maybe @Jegp wants to weigh in.

norse/torch/functional/spikes_to_times_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(
+        self,
+        spike_count: int = 1,
+        convert_indices_to_times: bool = True,
+        dt: float = 1e-3,
+    ):
+        """
+        Initialize decoder.
+        :param spike_count: number of elements (ascending) to return for each output channel.
```
cpehle: We use a different convention of specifying parameters in docs strings, could you please follow the convention based on what is done in other files.

norse/torch/functional/spikes_to_times_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+
+    def __init__(
+        self,
+        spike_count: int = 1,
+        convert_indices_to_times: bool = True,
+        dt: float = 1e-3,
+    ):
+        """
+        Initialize decoder.
```
Jegp: Do these docstrings generate ok? I haven't tried to split up module and __init__ function documentation like this before.

cpehle (changes requested):
LGTM up to the parameter doc string issue. Thanks for the contribution 🎉, @Jegp also has some input.

norse/torch/functional/spikes_to_times_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+            spike or if a spikes index wasn't returned in forward call.
+            """
+            (spike_indices, spk_rec) = ctx.saved_tensors
+            spikeidcs_size, batch_size, out_size = spike_indices.shape
+            noninf_spike_indices = spike_indices.flatten() != torch.inf
+
+            grad_local = torch.zeros_like(spk_rec, dtype=torch.float)
+
+            grad_local_indices = (
+                spike_indices.flatten().type(torch.int64)[noninf_spike_indices],
```
Jegp: Why `int64`?

norse/torch/functional/spikes_to_times_decoder.py
```diff
@@ -0,0 +1,98 @@
...
+                : spike_count
+                if spike_count < spike_input.shape[0]
+                else spike_input.shape[0]
+            ]
+            ctx.save_for_backward(spike_indices, spike_input)
+            ctx.shape = spike_input.shape
+            return spike_indices
+
+        @staticmethod
+        def backward(ctx, grad_output):
```
Jegp: Does this work with `torch.jit`?

Jegp (changes requested):
Nice work. Looks similar to how I did it in the other PR.

I added some minor comments, but most importantly I would appreciate it if we could
1. Test that the code works if the code is `jit`'ed
2. Add entries in the `__init__.py` files, so we can import the code directly via `norse.torch`. Specifically, you'd have to add it to https://github.com/norse/norse/blob/master/norse/torch/module/__init__.py

norse/torch/functional/spikes_to_times_decoder.py
lucablessing: I often used them like this, didn't lead to any problems yet

norse/torch/functional/spikes_to_times_decoder.py
lucablessing: When using `torch.int` for example, `IndexError: tensors used as indices must be long, byte or bool tensors` occurs, so I chose `int64`, but `long` could be used as well.

norse/torch/functional/spikes_to_times_decoder.py
lucablessing: Actually, I don't have any experience with `torch.jit` yet. When using on the `ToSpikeTimes` autograd function directly `torch.jit.trace` fails. When implementing something like
```
def to_spike_times_fn(spike_input: torch.Tensor, spike_count: torch.Tensor):
    return ToSpikeTimes.apply(spike_input, spike_count)
```
in `norse/torch/functional/to_spike_times_decoder.py` (like in `norse/torch/functional/threshold.py` with `HeaviErfc` and `heavy_erfc_fn` for example)  and using `torch.jit.trace(to_spike_times_fn, example_inputs=...)` it works, but with the warning
```
TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  spike_indices = torch.sort(indexed_spike_input, dim=0).values[:spike_count if spike_count < spike_input.shape[0] else spike_input.shape[0]]
```

Using `torch.jit.script` gives some errors, I'm currently reading into this, but maybe you can give me a faster note on how `torch.jit` is normally used here, i.e. in norse or when working with norse functions/modules.

norse/torch/functional/spikes_to_times_decoder.py
Jegp: Ah, you're using it as a lookup. Yep, that makes sense :)

norse/torch/functional/spikes_to_times_decoder.py
Jegp: Cool it works! I was hinting at creating a test for this, just so we know that this doesn't regress sometime in the future :-)

lucablessing: Okay :) So do you think the extra function `to_spike_times_fn` should be added as mentioned above, or leave as is in the latest commit?

cpehle: I think we can leave that addition to a separate commit.

Jegp:
Ah, one more thing. Wouldn't it make sense to put this in a `decode.py` file similar to the `encode.py` file?

norse/torch/functional/spikes_to_times_decoder.py
Jegp: Gotcha. I tried to generate the docs, and they're failing for some reason. None of the classes appear to have their own pages:
image

I'm happy to defer this to another issue - we need to clean up our documentation anyway :-)

norse/torch/functional/spikes_to_times_decoder.py
lucablessing: TorchScript seems to not support jit compilation on custom autograd function directly atm, but using on `to_spike_times_fn` seems to work fine.
So I can add a test for the Module only, which should suffice since it uses `ToSpikeTimes.apply` i.e. the autograd function, ensuring jitting when used inside the Module works.

lucablessing:
> Ah, one more thing. Wouldn't it make sense to put this in a `decode.py` file similar to the `encode.py` file?

@cpehle suggested to keep it in an extra file and import in something like `decode.py` in order to have authorship about that code/file in special. But since it seems you implemented something similar beforehand I don't know how much this applies here.

norse/torch/functional/spikes_to_times_decoder.py
Jegp: The lack of support can be solved by adding the annotation `@torch.jit.ignore` (https://github.com/norse/norse/blob/master/norse/torch/functional/superspike.py#L21). 

Thanks for seeing this through btw! :-)

Jegp:
> @cpehle suggested to keep it in an extra file and import in something like `decode.py` in order to have authorship about that code/file in special. But since it seems you implemented something similar beforehand I don't know how much this applies here.

I'm much less concerned about authorship than @cpehle. I trust you to decide what's best :-)

Jegp:
Lint it and LGTM! :-)

cpehle:
You can use ```black``` to reformat the code.

Jegp:
Thanks for the addition @lucablessing !

lucablessing:
You're welcome! Thanks for the helpful discussion regarding the changes.

PR #322, opened 2022-05-12
Build docker image only on push to master

cpehle:


Jegp (changes requested):
I would recommend that we split this into two workflows: one that only builds the container, but *without* pushing it (for PRs) and one that builds *and* pushes the latest version. That way, we can ensure that the container still builds for any submitted PRs.

cpehle:
> I would recommend that we split this into two workflows: one that only builds the container, but _without_ pushing it (for PRs) and one that builds _and_ pushes the latest version. That way, we can ensure that the container still builds for any submitted PRs.

I was under the impression that one of the first steps in the workflow was to login to docker hub. Even before the container is build. I don't know of a way to "build" the container without doing so.

Jegp:
There are some neat actions to achieve this, such as https://github.com/docker/setup-buildx-action

If you prefer to just disable it for now, we can also go ahead and open an issue later.

cpehle:
> There are some neat actions to achieve this, such as https://github.com/docker/setup-buildx-action
> 
> If you prefer to just disable it for now, we can also go ahead and open an issue later.

Yeah I think this should somehow be separate. I don't know if we need this action around at all, if buildx does similar checks.

Jegp (approved):

Issue #325, opened 2022-05-17
Investigate Changing Input Current Delay and Integration


cpehle:
LeonardBo brought up the following issue to our attention here (https://github.com/norse/norse/discussions/318):

> Input currently seems to affect the voltage at the 2nd timestep at the earliest.
> 
> I have notic,ed the reocurring pattern in my network that every LIF layer seemed to halt the forward progression of spikes in the network for a few timesteps. It didn't seem unnatural at first since the neuron's threshold may need several time steps to be overcome. However, even with very small thresholds the problem persisted so I dove depp into the code for the LIF step.
> 
> What I found is that in the very first timestep, the current input doesn't influence the voltage at all, as seen [here](https://github.com/norse/norse/blob/73d7c53cf0200fd1f1042e347102ebb09b630072/norse/csrc/lif.cpp#L28) -> it'll only affect the next time step's state.i. Now naturally, this effect will accumulate over several LIF layers; a reasonably sized network of 5 LIF activation layers will, at the very least, need 6 time steps to produce any meaningful output at all.
>
> Now since this is a hopefully intentional design decision I didn't open an issue but wanted to discuss it here. I can see that purely from the underlying ODE's point of view this might be fine. However, from a practical point of view that creates an unnecessary inefficient spiking behaviour.
>
> It seems to me, you might as well calculate i_new first and use that to update the neuron's voltage in [Line 16](https://github.com/norse/norse/blob/73d7c53cf0200fd1f1042e347102ebb09b630072/norse/csrc/lif.cpp#L16) instead of the old state's i. There might also be some other options to have the current input directly influence the current voltage. I might have missed some theoretical background justifying this, but it looks like that may fit the spiking model, too. Does that seem reasonable to you?

This issue is here to track potential solutions, including changing the behaviour outright. We need to think about whether we should address this concurrently with a move to the planned ODE integration interface.

Jegp:
I'm not sure I follow the logic here. I completely understand that the delay is "annoying" from a practical point of view. But it's also an inherent property of SNNs that they simply do not propagate signals immediately.

I'm wondering whether this is A) something Norse should consider and, if so, B) something we can include in the synapse dynamics. I'm saying that because we recently had talks about separating synapse and cell dynamics. That could be one way of getting about this issue, because we could have models without any form of delay.

cpehle:
[...]

Issue #89, opened 2020-07-16
Add tests for inference / training


cpehle:
By checking a small model trained on a simple task, it would become easier to test training and inference. It would also help test TorchScript support.

Jegp:
Is this closed with #123  @cpehle ?

cpehle:
> Is this closed with #123 @cpehle ?

The issue is a bit unspecific, but what I meant was a test which does end-to-end training on a toy dataset which doesn't blow up the CI budget.

Jegp:
How about using the odd-firing model in https://github.com/norse/notebooks/blob/master/parameter-learning.ipynb?
It's sufficiently complex to have to be learned and it shouldn't steal too many resources.

Issue #54, opened 2020-05-27
Create CI test to check learning


cpehle:
It would be useful to introduce a simple binary classification task, that could be used to test regressions in the different surrogate gradient implementations and modules.

PR #326, opened 2022-05-19
Feature no delay

cpehle:
This addresses #325, as far as I can tell there is no downside to this change. One thing that I noticed is that there are also some inconsistencies in how we handle resets and the naming conventions for current / spike integration. They should probably be adressed in a separate change.


norse/torch/functional/lif.py
```diff
@@ -392,8 +429,28 @@ def lif_feed_forward_integral(
         p (LIFParameters): parameters of a leaky integrate and fire neuron
         dt (float): Integration timestep to use
     """
-    z, v, i = norse_op.lif_super_feed_forward_integral(input_tensor, state, p, dt)
-    return z, LIFState(z=z, v=v, i=i)
+    outputs = []
```
Jegp, line 432: Why not use the cpp code? If you are concerned about parameterizing the surrogate gradient method, I suggest renaming the function to include `super` and then do the differentiation in the `LIF` module, such so we retain the performance benefits.

norse/torch/functional/lif.py
```diff
@@ -431,27 +488,10 @@ def _lif_feed_forward_integral_jit(
...
-        # compute reset
-        v_new = (1 - z_new) * v_decayed + z_new * p.v_reset
-        # compute current jumps
-        i_new = i_decayed + input_spikes
-
-        outputs.append(z_new)
-        state = LIFFeedForwardState(v=v_new, i=i_new)
-    return torch.stack(outputs), state
+    z, s = _lif_feed_forward_integral_jit(
+        input_tensor, state, LIFParametersJIT(*p), dt
```
Jegp, line 492: Circular reference.

Jegp (changes requested):
Thanks for the changes!

Before I can approve this, I still need to understand the motivation for this. Specifically, I would like some comparison that shows the need for this kind of change. Does it really provide the numerical stability we want? Does it actually solve the problem?

In principle, I think this can work, but this is also a big change in the fundamental workings of Norse. If we push this through, we need to make sure that all the tests are passing, notebooks are working, and the benchmarks are up to date.

norse/torch/functional/lif.py
cpehle: Yeah, I need to fix that.

norse/torch/functional/lif.py
cpehle: I don't think special casing like this is a good idea, right now. It breaks the expectation of the users and violates the interface.

Jegp: We already have those "special cases": https://github.com/norse/norse/blob/master/norse/torch/module/lif.py#L65

Jegp:
> This is really not about numerical stability. It is about whether spikes get integrated (in discrete time) immediately by the following layer, or not. As was pointed out in the discussion, the way we do it right now has the undesirable effect of accumulating latency if many layers are sequentially connected. 

Unfortunately, this means that the FF implementations will be inconsistent with the recurrent ones. I can imagine that situation in the short run, but I don't think it's desirable in the long run. For me consistency >> "undesirable" delays in deep layers. 
One possible path forward would be to accept this as status quo, but then decide on a solid separation of synapse and neuron dynamics that is consistent across models and layer types.

norse/torch/functional/lif.py
cpehle: I wouldn't say this is a special case, just a different algorithm. The choice of surrogate gradient function at the moment does nothing *only* in the case of LIF neurons. This is unexpected and not consistent with the documentation.

Jegp: Overall, I think it's desirable to abstract the gradient algorithms in the "method" parameter. The error in LIF neurons, where it's ignored is critically bad. But that doesn't mean we can't still have it. Perhaps I'll finally try to flesh out the Threshold #242  module in the weekend. That could be a good place to start discussing how the API could/should look longer term.

cpehle: But it can't be abstracted like this, what is missing is proper integration support. Similarly for the threshold module, if you want to pass things to C++ you do not want to support necessarily arbitrary callable, which is why I went for the enum change. Baring that the proper way to do it would probably to just pass a callable and eliminate the threshold function altogether.

cpehle:
I've now also tested that the mnist example notebook runs, with no degradation in performance. I think this change is good to go as is. I will expand the coverage back and we should try to integrate back the c++ code once it is sufficiently general.

Jegp (approved):
Despite the hit in performance after moving code from C++ -> Python, LGTM. Nice work :-)
And then I'll think about #89 / #54 so we can clearly measure where we are in terms of runtimes.

LennardBo:
I don't mean to resurrect past discussion, but I was wondering why weren't the non-recurrent lif cells not included int his change? The [`_lif_feed_forward_step_jit`](https://github.com/norse/norse/blob/6df3bff7feba5cf2269c4e944625558c14d89a8f/norse/torch/functional/lif.py#L328) which to my understanding has the non-recurrent lif step still features the old "laggy" implementation.

Sorry that I didn't bring up this comment before the merge despite raising the sisue to begin with; during that time I was preoccupied with other projects and then was waiting for an official release of a newer version with these issues resolved (ofc that may take a while).

PR #327, opened 2022-05-22
fix: Changed inplace spike latency encode to non-inplace

erikberter:
The functions spike_latency_lif_encode and spike_latency_encode in the functional modulo have in-place operations while applying the mask. I replace the boolean mask with an integer mask which tracks the aggregate of the spikes, which can be used with a comparator inside the torch.where.

I found the error while trying to build a model with an analogic convolutional neural network to extract features before a spiking classificator with time-to-first spike encoding. After the change was made, the video classificator worked fine.

### DIFF
--- a/norse/torch/functional/encode.py
+++ b/norse/torch/functional/encode.py
@@ -250,15 +250,16 @@ def spike_latency_lif_encode(
     """
     voltage = torch.zeros_like(input_current)
     z = torch.zeros_like(input_current)
-    mask = torch.zeros_like(input_current).bool()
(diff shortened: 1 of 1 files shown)


cpehle:
Thanks, this looks great. We would be happy if you could share the project you are working at some point!

cpehle (approved):
Thanks, this does enable flowing gradients through this layer aswell right?

erikberter:
> this does enable flowing gradients through this layer aswell right?

Indeed, I was able to fine-tune a ResNet-3D-18 through it.

PR #328, opened 2022-06-10
Feature spikes to times decoder

lucablessing:
Wrong implementation in `backward()` autograd function `ToSpikeTimes` caused and error (size mismatch) when using `spike_count` values larger than `1` and smaller than `spike_input.shape[0]`.

This is fixed by writing the `grad_output` values directly into the corresponding places (i.e. inject gradient at position of spike which corresponds to returned spike index in `forward()`) instead of first writing `1.0`s and then multiplying.

Also a test that the backward functionality works for all `spike_count` values is provided, since this wasn't covered before.

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,8 +1,9 @@
 name: Build docker images
 
 on:
-  pull_request:
-
+  push:
+    branches:
+      - master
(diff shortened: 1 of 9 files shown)


cpehle (approved):
LGTM. Potentially needs to be rebased though.

PR #329, opened 2022-06-12
doc: parameter learning

cpehle:


cpehle:
Yeah, although ideally I think these should be fixed in another pull request? The broken templates are not related to the parameter learning doc.

Jegp:
> Yeah, although ideally I think these should be fixed in another pull request? The broken templates are not related to the parameter learning doc.

Completely correct. Let me open another PR.

Jegp:
That should do it. I force-pushed it to get the history straight.

The commit should be attributed to you, but I can't seem to cheat GitHub to understand that you also pushed the commit. I am *not* trying to steal your work, so I would prefer if you could amend and force-push the commit again, @cpehle, to avoid ambiguities.

PR #330, opened 2022-06-15
Fixed sign in LIF equation documentation.

emijan-kth:


Jegp (approved):
Thank you for catching this!

PR #331, opened 2022-06-20
Feature ci benchmark

Jegp:
This PR adds automated CI benchmarking following each tags. Specifically, it

* Upon every tag, runs the benchmark script and appends the output to the `benchmark_results.csv` file
* Generates a new benchmark plot
* Shows a comparison of runtimes between versions


cpehle (approved):
Great :) especially that it runs self-hosted!

cpehle:
The doc build keeps failing, this should probably be adressed in a separate issue.

Jegp:
Thanks for the review! :-) 

> The doc build keeps failing, this should probably be adressed in a separate issue.

Agreed. I'll see to that and the binder container next.

PR #335, opened 2022-06-28
Fix[ci/spack] caching behavior

muffgaga:


muffgaga:
Hmm, I somehow thought that the restore-keys should be some kind of backup entries for the cache, but yes, using a symbolic name instead of the checksum would be less specific → I'll try it immediately :D.
(The extra restore-key `cache-main` should be more specific than `cache-`, i.e. as far as I understand the matching, this should not allow for "more" hits.)

Somehow I still feel this is related to non-existing `norse/main` cache entries — and we would only get those by running CI not only on PRs but "directly" on `main`.

muffgaga:
And yes, your suggested relaxation from `github.sha` to `github.ref` allows github cache to find entries from previous builds in the same PR (at least):

```
Run actions/cache@v2
…
Cache Size: ~739 MB (774712960 B)
…
Cache restored successfully
Cache restored from key: cache-97072da4d6f10a4775e0c21a4351245467627b98
```

BTW. by looking at the files again, I think that adding the "spack buildcache" didn't really change anything. I had already added a cached "persistent" [install folder](https://github.com/norse/norse/blob/main/.github/spack/config.yaml#L6), that now also works ;)… not sure what I originally tested there (maybe caching worked initially?).

```yaml
config:
#…
  install_tree:
    root: ~/.spack-cache/install
```

Jegp (approved):

muffgaga:
Jup, if builds on `main` would fill into something like `cache-main-latest` this would be perfect! (I'd keep the branch name in the cache key as there might be some need for another branch, and then it probably would cost some time to debug the wrong cache hit :p.)

And in terms of moving forward → sure, this PR is "complete" — caching is as "fixed" as it can be in terms of "ci/spack" :D

muffgaga:
@Jegp Or do you prefer to extend this PR here (and introduce another job that fills the main queue)?

Jegp:
> @Jegp Or do you prefer to extend this PR here (and introduce another job that fills the main queue)?
> 
> I'll just create another PR, and you can decide/select :D…

I don't have strong preferences to be honest. I'm more concerned about getting something functional and clean :-)

Now that you opened the other PR, let me squash this one. 

Thanks again for investing time into this!

PR #338, opened 2022-07-04
Cleanup[ci/spack]

muffgaga:
* the repo switched from `master` to `main` quite some time ago → follow up for spack
* use ninja for py-torch extension build (fixes a warning about this)
* unpins py-torch

### DIFF
--- a/.github/spack/build_spack_package.sh
+++ b/.github/spack/build_spack_package.sh
@@ -25,7 +25,7 @@ cp "$WORKSPACE/spack/package.py" custom_repo/packages/py-norse
 spack repo add "${TMP_DIR}/custom_repo"
 
 # we install a stripped down py-torch (no cuda, mpi, ...)
-PACKAGE_PYTORCH="py-torch@:1.10~cuda~mkldnn~rocm~distributed~onnx_ml~xnnpack~valgrind"
+PACKAGE_PYTORCH="py-torch~cuda~mkldnn~rocm~distributed~onnx_ml~xnnpack~valgrind"
 
 # the ubuntu CI runner runs on multiple cpu archs; compile for an old one
 ARCH="linux-ubuntu20.04-x86_64"
--- a/.github/spack/build_spack_package.sh
+++ b/.github/spack/build_spack_package.sh
@@ -39,16 +39,16 @@ spack compilers
 
 echo "spack spec of increasing specificity:"
 spack spec ${PACKAGE_PYTORCH}
-spack spec py-norse@master
-spack spec py-norse@master ^${PACKAGE_PYTORCH}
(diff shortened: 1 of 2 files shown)


cpehle:
Let me know, when I should take a look @muffgaga.

muffgaga:
> Let me know, when I should take a look @muffgaga.

It builds against py-torch@1.11.0 and the ninja warning seems gone → ready for review.

muffgaga:
BTW. nightly also ran for the first time.

As this would be very easy to do, it might be interesting to introduce matrix builds… different py-torch, python, gcc versions (e.g. at the "extremes" → verify oldest supported python (3.7?), py-torch (1.9?) and gcc (not sure?) and newest python, py-torch, gcc).
Opinions on this @Jegp @cpehle?

cpehle:
> BTW. nightly also ran for the first time.

Nice! 

> As this would be very easy to do, it might be interesting to introduce matrix builds… different py-torch, python, gcc versions (e.g. at the "extremes" → verify oldest supported python (3.7?), py-torch (1.9?) and gcc (not sure?) and newest python, py-torch, gcc). Opinions on this @Jegp @cpehle?

I think that would make sense, my only concern would be build times and making sure that reviews are not held up too much by very long ci / build times of those jobs.

cpehle (approved):
LGTM 👍.

Jegp:
Thanks for the work @muffgaga! That is, indeed, super nice :tada: 

Having matrix builds would be excellent for more dependencies, I think. I'm (slowly) moving towards something similar with the nix setup. But in principle, I don't have opinions *how* we achieve this -- as long as it doesn't add complexity to the Norse code itself.

muffgaga:
> > As this would be very easy to do, it might be interesting to introduce matrix builds… different py-torch, python, gcc versions (e.g. at the "extremes" → verify oldest supported python (3.7?), py-torch (1.9?) and gcc (not sure?) and newest python, py-torch, gcc). Opinions on this @Jegp @cpehle?

@cpehle:
> I think that would make sense, my only concern would be build times and making sure that reviews are not held up too much by very long ci / build times of those jobs.

With caching we are now down to O(4min) for the ci/spack build… the builds from the matrix can run in parallel, but I'm not sure about the available resources (which could limit parallelization again).

@Jegp:
> Having matrix builds would be excellent for more dependencies, I think. I'm (slowly) moving towards something similar with the nix setup. But in principle, I don't have opinions how we achieve this -- as long as it doesn't add complexity to the Norse code itself.

Complexity to the Norse source code would only be introduced if "more" entries in the build matrix shall be supported, e.g., code paths for backwards compatibility, or similar.

PR #339, opened 2022-07-05
Add[ci/spack] matrix build

muffgaga:


muffgaga:
@cpehle @Jegp I added oldest + newest in the range:
```
spack-python: ['python@3.7', 'python@3.10']
spack-pytorch: ['py-torch@1.9', 'py-torch@1.11']
```
for now (but not yet an "open interval" thing to detect future version mismatch).
We could also introduce `spack-compiler: [ 'gcc@10', 'gcc@11', 'clang@…' ]` for more coverage.

→ Renamed the caches again to explicitly mention the matrix selection → rebuild takes time :D.

muffgaga:
I had to skip the `python@3.10`/`py-torch@1.9` entry as this is not supported (`py-torchvision` does not support this particular Python version when constrained to a version supporting the given PyTorch version).

muffgaga:
BTW. adding a build option for the compiler would be easy; and on this runner image, there're at least some options (clang@10, 11 and 12; gcc@9.4.0 and the 10.3.0 that we currently use; i.e. unfortunately not fully state of the art) — yes/no/maybe :D?

cpehle (approved):
LGTM. Thanks :) happy that the CI story is getting better and better.

Issue #345, opened 2022-11-09
Implement BNTT (Batch Norm through time)


cpehle:
This appears to be an effective method to train larger scale convolutional spiking neural networks. The reference implementation can be found here: https://github.com/Intelligent-Computing-Lab-Yale/BNTT-Batch-Normalization-Through-Time. Effectively this just maintains a batch norm function for each timestep. I therefore suggest to implement an abstraction which allows us to "replicate" any kind of point wise applicable torch layer across the time dimension with independent parameters for each time step.

PR #348, opened 2022-11-09
feat: add replicate through time module

cpehle:
This feature will implement a general mechanism for replicating layers through time, in order to implement Batch Norm Through Time (BNTT) (#345).

### DIFF
--- a/norse/torch/module/replicate_through_time.py
+++ b/norse/torch/module/replicate_through_time.py
@@ -1,0 +1,14 @@
+import torch
+
+class ReplicateThroughTime(torch.nn.Module):
+    def __init__(self, module, T):
+        self.module_list = torch.nn.ModuleList([module for _ in range(T)])
+        self.T = T
+
+    def forward(self, x):
+        outputs = []
+        for ts in range(self.T):
+            y = self.module_list[ts](x[ts])
+            outputs.append(y)
+
+        return torch.stack(outputs)


norse/torch/module/replicate_through_time.py
```diff
@@ -0,0 +1,14 @@
+import torch
+
+class ReplicateThroughTime(torch.nn.Module):
+    def __init__(self, module, T):
```
Jegp, line 4: A brief docstring would be helpful

Jegp: Also, types would be neat :)

norse/torch/module/replicate_through_time.py
```diff
@@ -0,0 +1,14 @@
...
+        self.module_list = torch.nn.ModuleList([module for _ in range(T)])
+        self.T = T
+
+    def forward(self, x):
+        outputs = []
+        for ts in range(self.T):
+            y = self.module_list[ts](x[ts])
+            outputs.append(y)
+
+        return torch.stack(outputs)
```
Jegp, line 14: This will break if the module returns tuples (output, state).

Jegp (changes requested):
Nice addition!

I'm mostly concerned with how the `stack` will break if the module is recurrent (returns a tuple of output and state). A quick fix would be to distinguish on the return type. If you care about performance, you could do the check in the constructor.

Also, it would be nice to add things like docstrings and typing :-)

norse/torch/module/replicate_through_time.py
cpehle: Yes, but this isn't meant for tuples.

cpehle: Rather this is meant for modules which are applied "point wise" in time but with one module per timepoint

Jegp:
I see your point about not wanting this for stateful modules. I'd say let's go ahead. If you can add a few lines of documentation, that would be wonderfully helpful.

Jegp:
Closed for now. Contributions are not than welcome! ❤️

Issue #200, opened 2021-04-10
Validate export of ONNX graphs


Jegp:
To ensure optimal ONNX output, we should test that relevant loops are exported as loops. This is particularly relevant for the `SequentialState`, `Lift`, and recurrent models.

https://pytorch.org/docs/stable/onnx.html#tracing-vs-scripting

PMMon:
In order to validate the correct export of ONNX graphs for SNN models, I had a look into exporting a simple LIFCell to `onnx` format. 
It appears that the current LIFCell implementation, using **named** tuples result in errors, as it can be seen from the following trace: 

`RuntimeError: 

_lif_feed_forward_step_jit(Tensor input_tensor, __torch__.norse.torch.functional.lif.LIFFeedForwardState state, __torch__.norse.torch.functional.lif.LIFParametersJIT p, float dt=0.001) -> ((Tensor, __torch__.norse.torch.functional.lif.LIFFeedForwardState)):
Expected a value of type '__torch__.norse.torch.functional.lif.LIFFeedForwardState' for argument 'state' but instead found type 'Tuple[Tensor, Tensor]'`

Hence, I replaced relevant named tuples to simple tuples, which seems to solve this first problem as this output suggests: 

`graph(%input_tensor : Float(2, 3, strides=[3, 1], requires_grad=0, device=cpu),
      %1 : Long(1, strides=[1], requires_grad=0, device=cpu),
      %2 : Long(1, strides=[1], requires_grad=0, device=cpu)):
  %3 : Float(requires_grad=0, device=cpu) = onnx::Constant[value={0}]()
  %4 : Float(1, strides=[1], requires_grad=0, device=cpu) = onnx::Cast[to=1](%1)
  %5 : Float(1, strides=[1], device=cpu) = onnx::Sub(%3, %4) # /home/pmmon/.venvs/remote/lib/python3.8/site-packages/norse/torch/functional/lif.py:198:23
  %6 : Float(1, strides=[1], requires_grad=0, device=cpu) = onnx::Cast[to=1](%2)
  %7 : Float(1, strides=[1], device=cpu) = onnx::Add(%5, %6) # /home/pmmon/.venvs/remote/lib/python3.8/site-packages/norse/torch/functional/lif.py:198:23
  %8 : Float(requires_grad=0, device=cpu) = onnx::Constant[value={0.1}]()
[...]

PR #349, opened 2022-11-10
Added pytree state and support for torch.compile

Jegp:
This PR serves as place to discuss adding functionality to states. It's meant to partially address #181, improve usability, and slowly start having a crack at #200.

Concretely, I'm suggesting that we adds functions to states that lets us _almost_ treat them as tensors, while retaining support for Python's native Tuples as used in PyTorch's JIT'ing and ONNX.
The code will let us reformulate states as:

```python
from norse.torch.utils.pytree import StateTuple, MultipleInheritanceNamedTupleMeta
class LIFState(StateTuple, metaclass=MultipleInheritanceNamedTupleMeta):
    z: torch.Tensor
    v: torch.Tensor
    i: torch.Tensor
```

Which lets users do stuff like the following:
```python
s = LIFState(z=torch.as_tensor(0.0), v=torch.as_tensor(-65.0), i=torch.as_tensor(0.0))
s.to("cuda") # Move all tensors to cuda
s.int() # Convert all tensors to int
s.float() # Convert all tensors to float
s.broadcast(torch.empty(10, 20)) # Broadcast to shape of the given tensor
```

Note that this also works with default values and numeric members, although they won't be touched since they're outside the scope of PyTorch.

If there is support for this, we can roll it out to the states in the various models.

### DIFF
--- a/.github/workflows/benchmark_norse.yml
+++ b/.github/workflows/benchmark_norse.yml
@@ -12,7 +12,7 @@ jobs:
     # Only run on self-hosted Hugin machine
     runs-on: self-hosted
     container: 
-      image: python:3.10
+      image: python:3.11
       options: --gpus 1
     steps:
       - name: Checkout
--- a/.github/workflows/documentation_build.yml
+++ b/.github/workflows/documentation_build.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.9"
+          python-version: "3.11"
       - name: Install system dependencies
         run: sudo apt update && sudo apt install g++ -y
       - name: Fix sphinx-jupyterbook-latex bug
--- a/.github/workflows/documentation_publish.yml
+++ b/.github/workflows/documentation_publish.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11'
       - name: Install system dependencies
         run: sudo apt update && sudo apt install g++ -y
       - name: Fix sphinx-jupyterbook-latex bug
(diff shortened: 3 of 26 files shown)


Jegp:
Do you think this is worth-while to pursue @cpehle?

cpehle:
I would say generally speaking this is a good approach, as long as it is compatible with jit compilation. We should perhaps do a hackathon style pass through the codebase and identify all obstacles to jit compilation, then this would make sense to look at.

cpehle:
Overall I think this is a good approach, I've used something similar in JAX. Depending on the tree_math library. This lets you do things like

```python
from norse.base.types import ArrayLike
import dataclasses


@dataclasses.dataclass
@tree_math.struct
class SynapseState:
    a: ArrayLike
    b: ArrayLike

@dataclasses.dataclass
@tree_math.struct
class SynapseParameters:
    tau_r: ArrayLike
    tau_d: ArrayLike
    g_max: ArrayLike
    gamma: ArrayLike = 0.062
    e: ArrayLike = 0.0
    mg: ArrayLike = 0.0
    mg_dep: ArrayLike = 3.57

def dynamics(s: SynapseState, p: SynapseParameters):
    return SynapseState(a=-1 / p.tau_r * s.a, b=-1 / p.tau_d * s.b)
```

Naturally express the dynamics in terms of variables.

Jegp:
New changes includes support for `torch.compile` as a replacement for jit compilation for lif functions and modules. Note that I removed the default parameter values in the lif functional methods. It breaks the jit compilation, and I don't think it's a big problem to remove it because (1) users rarely use the functionals and (2) the modules still provide defaults.

The new pytree object is left in the tree, but breaks the dynamo compilation. But I suggest we merge it as-is and then build on it later.

Jegp:
The pytree NamedTuple compile issues are related to https://github.com/pytorch/pytorch/issues/91662 btw

Issue #350, opened 2022-11-30
Error in population_encode when encoding higher-dimensional Tensor


StrasserFlorian:
**Describe the bug**
Python throws the following error when i try to run population_encode on a 2D-Tensor of size [10000, 3] and out_features = 10:
`RuntimeError: The expanded size of the tensor (3) must match the existing size (10) at non-singleton dimension 2.  Target sizes: [10000, 10, 3].  Tensor sizes: [10]`
The problem is, that expanding the centres tensor without first unsqueezing it to the right shape throws an error. 

**To Reproduce**
1. create a 2D Tensor with `tensor_2d = torch.randn(10000, 3)`
2. try to run `norse.functional.population_encode(tensor_2d, 10)`
3. error happens

**Expected behavior**
I expected that population_encode would return a tensor of size `torch.Size([10000, 10, 3])`  with the probabilities required for the poisson encoding for each value in the original tensor.

**Screenshots**
![population_encode_error](https://user-images.githubusercontent.com/81424624/204857051-c599c141-779d-41c0-82f8-2171c10b6eea.PNG)

**Package versions:**
 - Python: 3.10.4 
 - Pytorch: 1.13.0.dev20220719

Jegp:
Thank you for the report @StrasserFlorian! This is definitely not expected behavior, and I can see there's a bug in the way we concatenate tensors if we get unexpected inputs with different dimensionality.

I was wondering, is the expected output really shaped `100, 10, 3` or is it `100, 3, 10`? That is, should the additional dimension be _inserted_ between the outer and inner dimensions, or is it desirable to append the additional dimension in case we wish to process higher-dimensional (>3) tensors.

StrasserFlorian:
Hi @Jegp,
i think your suggestion of appending the additional dimension at the end makes more sense. I only thought it might be on purpose to insert it in the middle, since that is how the size of the new Tensor is defined right now in the code:
` size = (input_values.size(0), out_features) + input_values.size()[1:]`

Jegp:
I've written a fix in this PR: https://github.com/norse/norse/pull/351

Note that this changes the location where the new dimension is inserted. I do think that the new version makes more sense. Always inserting at dimension 1 may be somewhat arbitrary. For example, if the input tensor shape is batch-width-height, it'll be inserted between batch-width, but if the input tensor is time-batch-width, it'll be between time-batch.

Would you mind checking out the PR code to see if that correctly addresses the problem? :-)

Jegp:
[...]

PR #351, opened 2022-12-04
Resolved shape error in population coding

Jegp:
Attempts to address #350


Jegp:
I'll merge this for now. There's a bug in the Windows installation, but we can track that in #354 .

Issue #355, opened 2022-12-19
Implement synops metric


Jegp:
Measuring the number of synaptic operations during the forward pass of an SNN would be a useful metric.

The code is probably straight-forward to implement with hooks similar to https://github.com/norse/norse/blob/main/norse/torch/module/sequential.py#L54

cpehle:
I actually have been asked about this a bunch of times. Would be still good to have it especially for a publication.

Jegp:
Fully agree. I guess the question is how to proceed. Just implementing "raw" synops metrics might be a little boring. And it's unclear to if that's helpful, mostly because I'm not sure "synops" is a commonly agreed-upon metric. Do you have an opinion here?
We could also lean on the work of others. I've been in touch with a PI from Stanford who recently developed some kind of toolbox/metric for this. Maybe that could be one approach?

PR #356, opened 2022-12-21
Added synops counter

Jegp:
Implements #355

### DIFF
--- a/norse/torch/module/__init__.py
+++ b/norse/torch/module/__init__.py
@@ -12,6 +12,7 @@ from .encode import (
     SpikeLatencyEncoder,
     SpikeLatencyLIFEncoder,
 )
+from . import hooks
 from .iaf import IAFFeedForwardState, IAFCell, IAFParameters
 from .leaky_integrator import LI, LICell, LILinearCell, LIParameters, LIState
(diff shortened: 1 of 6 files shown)


cpehle:
Do you know weather this. has a performance impact? I would rather solve this with something like a forward hook.

Jegp:
If the flag is not set, it'd be minimal because the initialization happens in the constructor. If the flag is set, however, I suspect it'll be costly. But, well. I guess that's the point :-)

Regarding using a forward hook for this, that may actually be possible, although the code would be slightly more convoluted. The way I could see this working would be to invert the control: offer a function that takes a module as input and sets the necessary buffers + hooks. Something like 

```def register_synops_hook(module: torch.nn.Module): ...

net = norse.torch.SequentialState(...)
register_synops_hook(net)
```

Would that be preferable?

cpehle:
I think so yes. This is not something that is so central to the function of the library that it should be hard coded into the control flow, but rather added on as needed. One way to do this would be some kind of forward hook or other "profiling" type operation.

Jegp:
Closed, c.f. https://github.com/norse/norse/issues/355

PR #357, opened 2022-12-29
Wip issue74 poisson encoder

theLamentingGirl:
rough.py is a dummy python file for my reference that can be ignored.

### DIFF
--- a/norse/torch/functional/encode.py
+++ b/norse/torch/functional/encode.py
@@ -134,7 +134,7 @@ def poisson_encode(
     seq_length: int,
     f_max: float = 100,
     dt: float = 0.001,
-    seed_generator: torch.Generator = None,
+    generator: torch.Generator = None,
     
 ) -> torch.Tensor:
     """
@@ -149,7 +149,7 @@ def poisson_encode(
         sequence_length (int): Number of time steps in the resulting spike train.
         f_max (float): Maximal frequency (in Hertz) which will be emitted.
         dt (float): Integration time step (should coincide with the integration time step used in the model)
-        seed_generator (torch.Generator): Generator for pseudorandom numbers. Usually, generator.manual_seed(seed value) is passed as the argument
+        generator (torch.Generator): Generator for pseudorandom numbers. Usually, generator.manual_seed(seed value) is passed as the argument
 
     Returns:
         A tensor with an extra dimension of size `seq_length` containing spikes (1) or no spikes (0).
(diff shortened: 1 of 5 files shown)


Jegp (changes requested):
Nice work @theLamentingGirl !

I would appreciate to hear your opinion on two suggestions below.

One is that the current `torch.manual_seed` sets the **global** seed. Which may be fine, but in some cases, that may not be what the user wants. What if they, for instance, just wants to set the seed in that function, but not globally?. One way around that is to use a [Generator](https://pytorch.org/docs/stable/generated/torch.Generator.html?highlight=generator#torch.Generator) which acts as the seed for the random sampling. This would be local, so we won't destroy whatever the user did outside this code. How about something like this?
```python
def poisson_encode(..., generator: Optional[torch.Generator] = None) -> torch.Tensor:
    ...
   return (torch.rand(..., generator=generator) ...)
```
This way, the user can either 1) set the general seed outside the function and _not_ use the generator (in which case it would be zero and [`torch.rand`](https://pytorch.org/docs/stable/generated/torch.rand.html?highlight=torch+rand#torch.rand) would get a `generator=None` argument - which is fine) or 2) specifically set the generator to the function which will be local _and_ device-specific (you can do `torch.Generator(device="cuda")` for example). Let me know your thoughts.

Regarding the second change, I would appreciate if you can remove the `rough.py` file so it doesn't gets pushed to the repo. Either by adding it to [`.gitignore`](https://git-scm.com/docs/gitignore) or by removing it entirely.

Finally, if you feel like nitpicking, it's good practice to write `seed is not None` instead of `seed != None` [because `!=` uses equality which can be overwritten](https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none). And in our case, that's bad because we don't want the user to accidentally shoot themselves in the foot by using an overwritten (hacked) `torch.Tensor` class. This is so hypothetical it likely won't happen in our lifetime, though :P

norse/torch/functional/encode.py
```diff
@@ -134,6 +134,8 @@ def poisson_encode(
     seq_length: int,
     f_max: float = 100,
     dt: float = 0.001,
+    seed_generator: torch.Generator = None,
```
Jegp, line 137: Could we call this parameter `generator` to follow the PyTorch conventions?

Jegp (changes requested):
Nicely done. I'm pretty happy with this, with the one exception that I think we should call the parameter `generator` instead of `seed_generator` to follow the PyTorch convention.

Otherwise, I think we should merge! :+1:

Jegp (approved):

Jegp:
Nice work! I'll leave the checks and merge it later. Thanks a lot @theLamentingGirl 🙏

Jegp:
Thanks a lot @theLamentingGirl :partying_face:

PR #363, opened 2023-03-08
fix: Improve CI build robustness

muffgaga:


muffgaga:
@cpehle you pinged me… I believe this is a caused by a broken CI build cache — the cleanest way would be to drop the CI build cache when things start to break; however, I can't come up with a reasonable (i.e. fast) solution ;).
→ Let's try to clean "some" more of the cached things, maybe it's already good enough.

cpehle:
Ok sounds good.

muffgaga:
Lint-step complains about files not touched in this commit:

> would reformat /home/runner/work/norse/norse/norse/dataset/speech_commands.py
> would reformat /home/runner/work/norse/norse/norse/torch/functional/test/test_encode.py
> would reformat /home/runner/work/norse/norse/norse/torch/module/conv.py
> would reformat /home/runner/work/norse/norse/norse/torch/module/snn.py

Finally the Spack CI job now fails in the build ;)… this seems to be related to the underlying gcc@12 compiler (cf. https://github.com/pytorch/pytorch/issues/77939)
```
2023-03-09T13:04:16.1834967Z   /usr/lib/gcc/x86_64-linux-gnu/12/include/avx2intrin.h:1155:32: error: 'mask_int32_v' may be used uninitialized [-Werror=maybe-uninitialized]
```

I could exclude building norse on py-torch@1.9 on a gcc@12 compiler and only try to build py-torch@1.11 or something?

.github/spack/build_spack_package.sh
```diff
@@ -42,7 +48,10 @@ if [ -n "${MATRIX_SPACK_PYTHON}" ]; then
 fi
 
 # the ubuntu CI runner runs on multiple cpu archs; compile for an old one
-ARCH="linux-ubuntu20.04-x86_64"
+ARCH="linux-ubuntu22.04-x86_64_v4"
```
muffgaga, line 51: Do we want to test multiple runners? This could go into the workflow file…

.github/workflows/spack_build.yml
```diff
@@ -10,6 +10,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: [3.9]
+        compiler-version: [gcc@11.3.0]
```
muffgaga, line 13: `gcc@12` doesn't build neither `py-torch@1.9` nor `py-torch@1.11`

muffgaga: related to https://github.com/pytorch/pytorch/issues/77939 which refers to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105593

cpehle (approved):
LGTM 🙏

.github/workflows/spack_build.yml
cpehle: Interesting and slightly strange.

.github/spack/build_spack_package.sh
cpehle: Might make sense but maybe not now. One big new things to test is M1 / Arm support.

cpehle:
> Lint-step complains about files not touched in this commit:
> 
> > would reformat /home/runner/work/norse/norse/norse/dataset/speech_commands.py
> > would reformat /home/runner/work/norse/norse/norse/torch/functional/test/test_encode.py
> > would reformat /home/runner/work/norse/norse/norse/torch/module/conv.py
> > would reformat /home/runner/work/norse/norse/norse/torch/module/snn.py
> 
> Finally the Spack CI job now fails in the build ;)… this seems to be related to the underlying gcc@12 compiler (cf. [pytorch/pytorch#77939](https://github.com/pytorch/pytorch/issues/77939))
> 
> ```
> 2023-03-09T13:04:16.1834967Z   /usr/lib/gcc/x86_64-linux-gnu/12/include/avx2intrin.h:1155:32: error: 'mask_int32_v' may be used uninitialized [-Werror=maybe-uninitialized]
> ```
> 
> ~I could exclude building norse on py-torch@1.9 on a gcc@12 compiler and only try to build py-torch@1.11 or something?~ 1.11 also fails on gcc@12… → pinning to `gcc@11.3.0`

Unsure how that comes about, should be fixed in new changeset then.

muffgaga:
Inching forward… and `py-torch@1.11.0` doesn't build with `gcc@11.3.0`, cf. https://github.com/pytorch/pytorch/issues/70297

cpehle:
Still appears to be a syntax error now @muffgaga.

PR #371, opened 2023-07-05
Adds NIR export

Jegp:
This PR adds support to export Norse model to the Neuromorphic Intermediate Representation format.


bauerfe:
Please excuse the question if this has been discussed already, but shouldn't snn-library-specific code be inside the respective libraries rather than here?

Jegp:
> Please excuse the question if this has been discussed already, but shouldn't snn-library-specific code be inside the respective libraries rather than here?

I very much agree with your statement that library-specific code should be inside the respective library. That way, the library will know what to do with the library, relieving NIR of that responsibility. 
But isn't that exactly what I'm doing here? This is a PR in the Norse project. So I believe "here" (aka Norse) is the correct place for Norse-specific code.

bauerfe:
For some reason I thought that this was the NIR repo. 🤦 I agree, it makes absolute sense here. Sorry for that.

Issue #369, opened 2023-06-12
CIFAR example not working


emijan-kth:
`python -m norse.task.cifar10`

results in the following error:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/raid/home/emil/norse/lib/python3.10/site-packages/norse/task/cifar10.py", line 193, in <module>
    parser = pl.Trainer.add_argparse_args(parser)
AttributeError: type object 'Trainer' has no attribute 'add_argparse_args'
```

4iar:
The `pl.Trainer.add_argparse_args` function was removed in Lightning 2.0 (https://github.com/Lightning-AI/lightning/pull/16708).

Downgrading to Lightning 1.x (`pip install pytorch-lightning==1.9.5`) should solve the issue for you

PR #372, opened 2023-07-17
Constrain pytorch-lightning version in example tasks README

4iar:
Lightning v2.0 was released recently and this is the default version installed with `pip install pytorch-lightning`

Unfortunately v2.0 removes the `pl.Trainer.add_argparse_args(parser)` function, which breaks the example tasks e.g. [cifar10.py](https://github.com/norse/norse/blob/main/norse/task/cifar10.py#L197) (#369)

This small PR updates the pytorch-lightning installation instructions to specify that pytorch-lightning v1 be used

### DIFF
--- a/norse/task/README.md
+++ b/norse/task/README.md
@@ -11,8 +11,8 @@ python -m norse.task.mnist
 ```
 
 Or, using [PyTorch Lightning](https://pytorchlightning.ai/) to scale to multiple GPUs and simplify a lot of the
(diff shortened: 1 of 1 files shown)


Jegp (approved):
Thank you for the correction :-)

Jegp:
Ideally, we should also update the examples. Using the newest version of lightning could be neat.

Would you be interested in helping out strengthening the examples 4iar? Apart from the present examples, I have some code that may be interesting to supplement the existing notebooks with.

PR #376, opened 2023-08-30
Feature rf

Jegp:
Updated temporal scales according to [Lindeberg (eq. 20)](https://link.springer.com/article/10.1007/s10851-015-0613-9).


norse/torch/functional/receptive_field.py
```diff
@@ -158,9 +168,21 @@ def spatial_receptive_fields_with_derivatives(
...
+    r"""
+    Provides temporal scales according to [Lindeberg2016].
+    The scales will be logarithmic by default, but can be changed by providing other values for c.
+    .. math:
+        \tau_k = c^{2(k - K)} \tau_{max}
+        \mu_k = \sqrt(\tau_k - \tau_{k - 1})
+    Arguments:
+      n_scales (int): Number of scales to generate
+      min_scale (float): The minimum scale
+      c (float): The base from which to generate scale values. Should be a value between 1 to 2
```
dimitriskor, line 183: Following the pattern from the previous sampling distributions (min_val, max_val, num_of_values), shouldn't the scales be selected from min and max values instead of the base c? [Lindberg 3.4: _Logarithmic Distribution of the Temporal Scales with Given Minimum Scale_](https://link.springer.com/article/10.1007/s10851-015-0613-9)

dimitriskor (commented):
Taus selection over time-scale

norse/torch/functional/receptive_field.py
Jegp: I can see this making sense in two ways: either to have a min/max setting, as you describe, or to provide `min_scale` and `c` to automatically spread out the scales. Perhaps we should allow both?

Jegp: What do you think about the current solution?

norse/torch/functional/receptive_field.py
```diff
@@ -158,9 +168,28 @@ def spatial_receptive_fields_with_derivatives(
...
+    Arguments:
+      n_scales (int): Number of scales to generate
+      min_scale (float): The minimum scale
+      max_scale (Optional[float]): The maximum scale, given the growth parameter c. Defaults to 1.
+      c (Optional[float]): The base from which to generate scale values. Should be a value between 1 to 2 exclusive. Defaults to sqrt(2).
+
+    .. [Lindeberg2016] Lindeberg 2016, Time-Causal and Time-Recursive Spatio-Temporal Receptive Fields, https://link.springer.com/article/10.1007/s10851-015-0613-9.
+    """
+    ks = torch.arange(min_scale, min_scale + n_scales + 1)
+    taus = c ** (2 * (ks - n_scales)) * max_scale
```
dimitriskor, line 194: Could someone provide both c and max_scale? I believe that in case where max_scale is provided, c should be calculated within the function with the formula $\ c = (\frac{max\textunderscore scale}{min\textunderscore scale})^{\frac{1}{2(n\textunderscore scales-1)}}$

dimitriskor (approved):

PR #377, opened 2023-09-12
Removed deprecated workflows

Jegp:
Prepare for publishing


cpehle:
I would like to keep the spack build somehow, we should probably not run it on every pull request. But it is useful to check that it is compatible.

Jegp:
> I would like to keep the spack build somehow, we should probably not run it on every pull request. But it is useful to check that it is compatible.

Thanks for the input! :)

That would be wonderful. I was worried it's not maintained anymore. Would you be ok with merging this and then work on fixing the workflow in a separate PR?

PR #394, opened 2024-04-26
fix: current code depends (at least) on torch >= 2.2

muffgaga:
When trying `import norse` in the latest RC of the EBRAINS Software Distribution (24.04-RC), I get:

```
ImportError: cannot import name 'register_pytree_node' from 'torch.utils._pytree' (/srv/main-spack-instance-2402/spack/var/spack/environments/experimental/.spack-env/view/lib/python3.8/site-packages/torch/utils/_pytree.py)
```

This seems to be caused by a wrong dependency statement in terms of torch.
The used function has only been introduced in torch 2.2.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dependencies = [
   "numpy",
-  "torch>=2.0.0",
+  "torch>=2.2.0",
   "torchvision>=0.15.0",
   "nir",
   "nirtorch",


Jegp (approved):
Thanks for the addition, @muffgaga!

Would you like a new release for this, so you can install it without having to pin the (pip) install to a branch?

muffgaga:
> Thanks for the addition, @muffgaga!
> 
> Would you like a new release for this, so you can install it without having to pin the (pip) install to a branch?

In terms of the EBRAINS Software Distribution it's fine, I'll just update the spack dependencies (and add a comment that for `v1.1.0` the project file is just slightly wrong ;) ).

→ I.e. no need to adjust your release schedule/roadmap.

PR #395, opened 2024-06-25
fix for Lift of LICell

dimitriskor:
This is a fix for the error that occurs when performing Lift() in LICell() modules:
```
TypeError: zeros() received an invalid combination of arguments - got (dtype=torch.dtype, device=torch.device, ), but expected one of:
 * (tuple of ints size, *, tuple of names names, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (tuple of ints size, *, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
```

### DIFF
--- a/norse/torch/module/leaky_integrator.py
+++ b/norse/torch/module/leaky_integrator.py
@@ -85,7 +85,7 @@ class LICell(SNNCell):
         state = LIState(
             v=self.p.v_leak.detach(),
             i=torch.zeros(
-                *input_tensor.shape,
+                input_tensor.shape,
                 device=input_tensor.device,
                 dtype=input_tensor.dtype,
             ),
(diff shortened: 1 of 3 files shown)


Jegp:
Excellent addition, thank you @dimitriskor!

Can you quickly run the changes through the black formatter? That's the only failing test. It should be easy to do with a few commands from the `norse` directory:

```bash
pip install black 
black .
```

Jegp:
Nicely done sir, thank you!

There is still one test failing, but I assume this has to do with the numpy 2.0 thing. That should be treated in a different PR, I would think.

Jegp (approved):

PR #397, opened 2024-07-09
fix transfer sampled and parametrized rf modules to cuda

dimitriskor:
ParameterizedSpatialReceptiveField2d and SampledSpatialReceptiveField2d where not transferring submodule's weights  to cuda.

Fixed by registering the weights as buffers.

### DIFF
--- a/norse/torch/module/receptive_field.py
+++ b/norse/torch/module/receptive_field.py
@@ -97,7 +97,9 @@ class SpatialReceptiveField2d(torch.nn.Module):
                 )
                 if self.aggregate:
                     self.out_channels = fields.shape[0]
-                    self.weights = fields.unsqueeze(1).repeat(1, self.in_channels, 1, 1)
+                    self.register_buffer(
+                        "weights", fields.unsqueeze(1).repeat(1, self.in_channels, 1, 1)
+                    )
                 else:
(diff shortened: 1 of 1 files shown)


Jegp (approved):
That's a great addition @dimitriskor, thank you :tada:

norse/torch/module/receptive_field.py
```diff
@@ -112,7 +114,9 @@ def _update_weights(self):
                         in_weights = empty_weights.clone()
                         in_weights[i] = fields
                         weights.append(in_weights)
-                    self.weights = torch.concat(weights, 1).permute(1, 0, 2, 3)
+                    self.register_buffer(
```
Jegp, line 117: @dimitriskor, I just realized that buffers are not considered a model parameter. Should this have been `register_parameter`? https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module

dimitriskor: The decision for declaring the weights as buffers has been deliberate, as they are only part of the computational graph and not leafs.
The grad is set to True and the gradient computation flows backwards to the rf parameters which are the leaf nodes of the computational graph.

PR #398, opened 2024-07-10
Receptive field position

dimitriskor:
Added the position (x, y) of the receptive field in the Cartesian plane as optimizable  parameters.

### DIFF
--- a/norse/torch/functional/receptive_field.py
+++ b/norse/torch/functional/receptive_field.py
@@ -7,13 +7,17 @@ from typing import List, Tuple, Union, Optional
 import torch
 
 
-def gaussian_kernel(size: int, c: torch.Tensor, domain: int = 8) -> torch.Tensor:
+def gaussian_kernel(
+    size: int, c: torch.Tensor, x: torch.Tensor, y: torch.Tensor, domain: int = 8
+) -> torch.Tensor:
     """
     Efficiently creates a differentiable 2d gaussian kernel.
 
     Arguments:
       size (int): The size of the kernel
       c (torch.Tensor): A 2x2 covariance matrix describing the eccentricity of the gaussian
+      x (torch.Tensor): The receptive's field center position in x-axis
+      y (torch.Tensor): The receptive's field center position in y-axis
       domain (int): The domain of the kernel. Defaults to 8 (sampling -8 to 8).
     """
     ci = torch.linalg.inv(c)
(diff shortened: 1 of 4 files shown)


norse/torch/functional/receptive_field.py
```diff
@@ -7,20 +7,26 @@
...
+      x (torch.Tensor): The receptive's field center position in x-axis
+      y (torch.Tensor): The receptive's field center position in y-axis
       domain (int): The domain of the kernel. Defaults to 8 (sampling -8 to 8).
     """
     ci = torch.linalg.inv(c)
     cd = torch.linalg.det(c)
     fraction = 1 / (2 * torch.pi * torch.sqrt(cd))
     a = torch.linspace(-domain, domain, size)
     xs, ys = torch.meshgrid(a, a, indexing="xy")
+    xs = xs - x
```
Jegp, line 28: This effectively means that a positive x leads to a left-shift. Is that on purpose?
According to the docs, the I would assume a positive x shifts the RF to the right.

norse/torch/module/receptive_field.py
```diff
@@ -157,16 +160,22 @@ def __init__(
         scales: torch.Tensor,
         angles: torch.Tensor,
         ratios: torch.Tensor,
+        x: torch.Tensor,
```
Jegp, line 163: Would it make sense to default the x, y coordinates to 0? I assume most people want centered RFs. Or?

norse/torch/module/receptive_field.py
```diff
@@ -229,15 +247,19 @@ def __init__(
...
```
Jegp, line 250: See previous comment.

Jegp (commented):
Very nice work here @dimitriskor! You've added a robust change with great tests to avoid regressions. I would perhaps suggest to have a default parameter value for the x,y coordinates *in the class-based API*. What do you think?

norse/torch/functional/receptive_field.py
dimitriskor: A positive x leads to a left-shift of the domain, which results in the kernel shifting to the right in the transformed domain. 
A value x<sub>0</sub> performs the transformation g(x) --> g(x-x<sub>0</sub>)

norse/torch/module/receptive_field.py
dimitriskor: I agree with your comment. I will add the default to 0

norse/torch/functional/receptive_field.py
Jegp: Ah, I see. That actually makes sense, thank you!

Jegp (approved):
Nice work and follow-up @dimitriskor. LGTM!

PR #399, opened 2024-07-16
added rf position default values

dimitriskor:
added rf position x, y default values to 0.0


Jegp (approved):
Nice work! This will make the interface easier to use I reckon

PR #401, opened 2024-08-07
Added log update option in RFs

dimitriskor:
Added log update option for scale and ratio in RFs

### DIFF
--- a/norse/torch/module/receptive_field.py
+++ b/norse/torch/module/receptive_field.py
@@ -51,14 +51,51 @@ class SpatialReceptiveField2d(torch.nn.Module):
         aggregate: bool = True,
         domain: float = 8,
         optimize_fields: bool = True,
+        optimize_log: bool = False,
         **kwargs,
     ) -> None:
         super().__init__()
+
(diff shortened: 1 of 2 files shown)


norse/torch/module/receptive_field.py
```diff
@@ -51,14 +51,51 @@ def __init__(
         aggregate: bool = True,
         domain: float = 8,
         optimize_fields: bool = True,
+        optimize_log: bool = False,
```
Jegp, line 54: Shouldn't this default to true?

norse/torch/module/receptive_field.py
```diff
@@ -179,12 +234,32 @@ def __init__(
         optimize_ratios: bool = True,
         optimize_x: bool = True,
         optimize_y: bool = True,
+        optimize_log: bool = False,
```
Jegp, line 237: Again, isn't the sensible default `True`?

dimitriskor: It's true that setting it to true might make more sense (leading to a more stable training).

PR #402, opened 2024-08-13
RF log update fix

dimitriskor:
Fix error due to default optimize_log set to True
Fix device mismatch error on backward pass by setting the has_updated parameter in the RF __init__() functions to True


Jegp (approved):
All this looks good, LGTM!

There is a test failing in the pytree code. But that's probably not be related to this PR, so feel free to merge and I can look into the problem in a separate PR

PR #416, opened 2025-02-17
Adding bio parameter

KhalifeHilal:
Here I added the bio parameter from nest to norse like we discussed it.


Jegp (approved):
Nice changes! Thank you for adding this :-)

PR #417, opened 2025-03-10
docs: update parameters page to use v1.1.0 interface

BorjaEst:
Updates some issues on the documentation page:
https://norse.github.io/norse/pages/parameters.html
Check style before merging.

### DIFF
--- a/docs/pages/parameters.ipynb
+++ b/docs/pages/parameters.ipynb
@@ -7,19 +7,20 @@
     "(page-parameter-learning)=\n",
     "\n",
     "# Parameter learning in Spiking Neural Networks\n",
+    "\n",
     "Author: Christian Pehle\n",
     "\n",
-    "As we have already seen in {ref}`page-spiking` Neuron models come with parameters, such as membrane time constants, which determine their dynamics. While those parameters are often treated as arbitrarily treated constants and therefore *hyperparameters* in a machine learning context, more recently it has become clear that it can be benefitial to treat them as *parameters*. In this notebook, we will first learn how to initialise a network of LIF neurons with distinct, but fixed time constants and voltage thresholds and then how to incorporate them into the optimisation. As we will see, this is largely facilitated by the pre-existing ways of treating parameters in PyTorch."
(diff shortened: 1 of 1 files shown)


jeremyforest:
I've tried running your code locally and my LIFParameters don't change, i.e. I can't reproduce the output graphs. Any insight ?

BorjaEst:
hello @jeremyforest 
What do you mean by LIFParameters don't change?
If you mean e.g.:
`np.histogram(norse.torch.functional.lif.LIFParameters().tau_mem_inv + 20*torch.randn(100))`
I do not think it is possible to reproduce unless the seed is fixed, but there was no seed indicated on the previous docs.

`ParametrizedLIFRecurrentCell` also has random elements when setting the internal wegihts, so to get the exact same previous soulution might not be possible.

Maybe an easy alternative might be to simply update the cells of the import statements to the correct ones (as it is for documentation purposes) but it is indeed to cheat a bit.

BorjaEst:
Now I see, from the original it does not seem the gradient was to strong, the difference in the change was very small.
So it might be actually 2 things:
 - Gradient is near 0 in that direction, so it does not change the optimal membrane potential for this problem. 
 - Any update in the gradient propagation in the last versions? Do you remember with which version was the documentation executed with the correct results? I do not see any recent changes that might influence on the last changelogs.

I have executed the notebook once more but results are similar. I have also tried `tau_mem_inv_after - tau_mem_inv_before` and it provides small value differences, but all very small compared to the values. 

```
array([ 0.04637909,  0.07762146,  0.13432312,  0.06166077,  0.0456543 ,
        ...
        0.06943512,  0.05469513,  0.17555237, -0.04631805, -0.05114746],
      dtype=float32)
````

Unfortunatelly I cannot be of much more help.

Jegp:
@jeremyforest have you tried adding the LIFParameters as a [torch Parameter](https://pytorch.org/docs/stable/generated/torch.nn.parameter.Parameter.html)? By default, the neuron parameters are not being optimized.

That would actually be a neat feature to have. snnTorch, for instance, has a flag that's called something like  `learn_parameters` which would toggle the training of parameters by assigning them as a torch Parameter.

Jegp:
Thank you for this update @BorjaEst! Super helpful and way overdue :pray:

Jegp (approved):

BorjaEst:
Thanks to you for this great tool!

PR #423, opened 2025-07-07
Update to latest NIR and NIRTorch

fabio-innatera:


cpehle (changes requested):

fabio-innatera:
@cpehle Are you OK with untangling the code reformatting from other changes into separate commits, or should we do separate PRs completely?
Happy with both.

Would probably be three PRs:
* One the GitHub build system fixes + Python version upgrade (Python 3.8 is no longer supported, but still in the GitHub workflow and therefore failing) + code reformatting
* One for the v_reset + type_checking
* One with the new NIRTorch API usage

fabio-innatera:
I've done some heavy git surgery and we should have nice separate commits now.

I am not sure about squashing the commits, after the clean up they all fix quite separate things.
But I can do that if you think it makes sense.

I think there are two reasons for the code formatting:
a) Newer black version, 25.1.0 vs 22.10.0 (I checked, this alone only explains like 3 line changes)
b) CI being broken and not actually checking for code formatting, so PRs were merged that had incorrect formatting

cpehle:
There just shouldn't be whitespace changes in files that otherwise haven't been touched. Otherwise I am fine with what @Jegp proposed.

cpehle (approved):

PR #424, opened 2025-07-09
NIR Leaky Integrator node import

fabio-innatera:
We can already export LIBoxCells, but not import.

The PR is against the update-nirtorch branch, because I think it makes more sense to use the new NIRTorch API
support from there.


Jegp:
Nice initiative. I like that you added a test for it. It's a little hard to separate the additions here from the other branch. What about waiting for the other branch to merge and then rebase this so we can add it in as a separate commit? Alternatively, we can merge it in and squash it all together. Preferences?

PR #426, opened 2025-07-14
Fix CI and make tests deterministic

fabio-innatera:


Jegp:
This is the CI part of #423

PR #439, opened 2026-03-04
Add more flexibility for exporting and bypassing of custom modules

mneher11:
Currently, only mappings defined inside Norse are valid for exporting `torch.nn` models to NIR. However, custom variations of Norse neuron implementations can exist that should still map to a specific NIR primitive. This pull request adds two contributions:

### Support for mapping custom modules to NIR primitives
The `to_nir` function has been extended to accept a `custom_mapping` parameter. This parameter is a dictionary mapping `torch.nn.Module` subtypes to NIR primitives and enables users of Norse to specify additional mappings from custom neuron implementations. It gets combined with the `mapping_dict` defined by Norse. Additionally, a `custom_stateful_modules` analogous to the `stateful_modules` set by Norse has been added to ensure correct tracing of custom stateful modules.

### Support for bypassing no-op and transparent modules
This change depends on neuromorphs/NIRTorch#50, and targets the selective bypassing of user-specified modules. Modules like `torch.nn.Dropout` or `torch.nn.Identity` do not contribute to the computation at runtime, and can therefore be viewed as transparent. This change exposes the functionality added in the NIRTorch pull request in Norse, enabling uses to bypass selected modules during tracing.

### DIFF
--- a/norse/torch/utils/export_nir.py
+++ b/norse/torch/utils/export_nir.py
@@ -122,7 +122,6 @@ def to_nir(
     model_name: str = "norse",
     time_scaling_factor: float = 1,
     custom_stateful_modules: typing.Set[typing.Type[torch.nn.Module]] = set(),
-    custom_bypass_modules: typing.Set[typing.Type[torch.nn.Module]] = set(),
     custom_mapping: typing.Dict[torch.nn.Module, typing.Callable[[torch.nn.Module], nir.NIRNode]] = {},
     type_check: bool = True,
 ) -> nir.NIRNode:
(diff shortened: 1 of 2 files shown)


Jegp (approved):
This is an excellent addition, thank you @mneher11. 

You've neatly added tests and docstrings. I'm in principle happy to approve this, but do you know why the tests fail? I can't access the logs... Could you try to re-run them? Either by re-triggering (I can't seem to do that either) or pushing a mock commit (that we can squash anyway)?

mneher11:
I re-triggered the GitHub actions runner with a simple comment commit.
Two of the failing tests seem to be related to #448 and should be fixed in that PR. The last failing test is related to the bypass changes:
```
FAILED norse/torch/utils/test/test_nir_export.py::test_bypass_module - AssertionError: assert 7 == 6
```
This is due to the NIRTorch package not being updated on PyPi (the GH runner automatically pulls nirtorch==2.6, which does not have the support for bypassing modules yet even though that commit was already merged in neuromorphs/NIRTorch#50)

PR #444, opened 2026-06-14
feat: implement synaptic operations (SynOps) metric (#355)

### LINKED ISSUE #355: Implement synops metric

TLXyloph:
## Summary

Implements the SynOps (synaptic operations) metric requested in #355.

SynOps counts the number of multiply-accumulate events triggered by spikes passing through weight layers during the forward pass — a standard metric for quantifying computational cost and energy efficiency in spiking neural networks.

The design mirrors the existing `RegularizationCell` / `regularize_step` pattern:
- **`norse/torch/functional/synops.py`** — pure functions (`linear_synops_accumulator`, `conv_synops_accumulator`, `synops_accumulator`) for counting synops given a spike tensor and a weight layer
- **`norse/torch/module/synops.py`** — `SynOpsCounter` module that wraps an `nn.Linear` or `nn.Conv2d` and accumulates synops across forward steps, with a `reset()` method for resetting between epochs
- Exports added to `norse/torch/__init__.py`

### Formulas used
- `nn.Linear(in, out)`: `synops += z.sum() * out_features`
- `nn.Conv2d`: `synops += z.sum() * out_channels * kH * kW / groups`

### Usage example
```python
import torch
from torch import nn
import norse.torch as snn

cell = snn.LIFCell()
counter = snn.SynOpsCounter(nn.Linear(2, 8))

data = torch.ones(5, 2) + 10  # drives all neurons to fire
z, state = cell(data)
z, total_synops = counter(z)
print(total_synops)  # synops accumulated so far

counter.reset()      # reset between epochs
```

## Test plan
- [x] 11 tests in `norse/torch/module/test/test_synops.py`
  - Functional: `nn.Linear` (with/without initial state, zero spikes)
  - Functional: `nn.Conv2d` (standard and grouped convolutions)
  - Functional: dispatch via `synops_accumulator`
  - Module: single-step, multi-step accumulation, state persistence, reset
- [x] Full test suite: 681 passed, 0 new failures
- [x] `black --check` passes on all new/modified files
- [x] `pylint` scores 10.00/10 on all new files

Closes #355


Jegp (changes requested):
I like the approach: use functional blocks to count SynOps and fail if you encounter a layer that's not supported. That's clear for the user, nice.

My only point would be to follow the conventional doc strings. I'm mostly saying that so it'll render correctly in the docs -- if you can convince me that this format also works, I'm fine to go ahead :)

Thanks for the PR in any case!

cpehle:
I would also request to remove the claude coauthor attribution in the commit. At the end of the day you are fully responsible for the commit and not the tool you used.

TLXyloph:
Will do, Mr. Pedersen and Dr. Pehle.

I apologize; I'm very new to open-source contribution and was focused on trying to get the PR to pass tests, but I see how I should do the docstrings now.

I also do see your point, Dr. Pehle. I will turn Claude coauthorship off for the foreseeable future, and remove Claude coauthor attribution from my PRs.

Thank you!

Jegp (approved):
Very good. Thank you for incorporating our suggestions and for pushing this through!

PR #448, opened 2026-07-20
Fix ONNX export with integer parameter tensors (torch 2.13 dynamo exporter)

### LINKED ISSUE #447: ONNX export fails on torch 2.13 new exporter — DispatchError for aten.mul.Scalar
### Summary
The ONNX export tests (`norse/torch/test/test_onnx.py::test_export_onnx_li` and `::test_export_onnx_lif`) fail on recent torch, where `torch.onnx.export` uses the new dynamo-based exporter. Export aborts with a `DispatchError`/`ConversionError` rather than producing a model.

### Environment
- `torch` 2.13.0
- `onnxscript` 0.7.1
- Python 3.11
- norse `main`

### Reproduction
```python
import torch
import norse.torch as snn

p = snn.LIParameters(tau_syn_inv=torch.as_tensor(42))   # int64 scalar
net = snn.SequentialState(snn.LICell(p))
torch.onnx.export(net, torch.randn(2, 1, 10), "snn_li.onnx")
```

### Error
```
torch.onnx._internal.exporter._errors.DispatchError:
  No ONNX function found for <OpOverload(op='aten.mul', overload='Scalar')>.
  Failure message: No decompositions registered for the real-valued input
... -> ConversionError: Failed to convert the exported program to an ONNX model (step 3/3)
```

### Analysis
The graph lifts `tau_syn_inv` as an `INT64` constant, so the multiply is traced as `aten.mul.Scalar` (int overload), which the new exporter has no registered ONNX function for. Passing a **float** tensor (`torch.as_tensor(42.0)`) keeps the constant `FLOAT` and the export succeeds — so it is dtype-triggered, but the root cause is norse's export path relying on an op overload the new `torch.onnx` exporter cannot dispatch.

### Question for maintainers
Does norse intend to support torch 2.13 / the new dynamo ONNX exporter? If so, is the preferred fix (a) narrow — ensure parameter tensors are float in the export tests/paths — or (b) broader — audit ONNX-export compatibility and bump the supported torch range? Happy to open a PR once you have scoped it.

JPL11:
Fixes #447.

## Root cause

The LI/LIF dynamics compute `dt * p.tau_x_inv * expr` left to right, so an integer-dtype parameter tensor (as constructed in `test_onnx.py`: `tau_syn_inv=torch.as_tensor(42)`) is first multiplied by the python scalar `dt`. In the exported program that materializes as a multiply of an int64 tensor by a scalar, and torch 2.13's dynamo-based exporter has no ONNX dispatch for `aten.mul.Scalar` with an integer input, so export aborts with the `DispatchError`/`ConversionError` from the issue.

## Fix

Reassociate to `dt * (p.tau_x_inv * expr)` in `leaky_integrator.py` and `lif.py`: the parameter tensor first multiplies the floating-point state expression, promoting to float before any scalar multiply ever happens. Purely a reassociation; the arithmetic is unchanged.

## Verification

- `test_onnx.py` plus the LI/LIF functional tests pass (12 passed), confirming numerics are unchanged.
- Inspecting `torch.export.export` on the issue's reproducer: before this change the graph contains `aten.mul.Tensor(int64, -0.001)` (which the 2.13 exporter rejects as `mul.Scalar`); after it, the graph contains no integer-times-scalar multiply at all, so the failing dispatch can no longer be reached.
- Note: I verified on torch 2.12 (the exporter accepts the fixed graph there too); I could not run 2.13 locally, so CI on a newer torch is the final confirmation.

The same `dt * p.tau_x_inv * expr` pattern exists in other neuron models (`lif_adex`, `lif_ex`, `izhikevich`, ...). Happy to apply the same reassociation across those in this PR or a follow-up if you'd like.


cpehle:
Good catch, definitely would make sense to apply this fix uniformly.

JPL11:
Applied reassociation in all remaining neuron models: lif_ex, lif_adex, lif_box, lsnn, coba_lif, lif_mc_refrac, leaky_integrator_box izhikevich and correlation_sensor.

